PR #68, opened 2015-07-02
Fix zero pwd

jreidinger:
manually tested and works. Empty root password is not problem as it is not allowed ( also tested ).

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jul  2 11:37:00 UTC 2015 - jreidinger@suse.com
+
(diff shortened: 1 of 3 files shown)


package/yast2-users.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jul  2 11:37:00 UTC 2015 - jreidinger@suse.com
+
+- allow root password to be set 0 (zero) (bnc#930909)
```
mvidner: > Allow setting the root password to "0" (the digit zero) (bnc#930909)

Otherwise LGTM

jreidinger:
thanks

PR #71, opened 2015-07-08
Add basic CheckUser tests

imobach:
Add basic test for Users.CheckUser function.


lslezak:
At first I wanted to ask you to rewrite it into RSpec but then I realized that it's Perl and that's not possible... 

So LGTM.

imobach:
I had the same though :)

Thanks!

PR #72, opened 2015-09-01
Move users creation in AutoYaST to finish client

imobach:
It's related to https://github.com/yast/yast-autoinstallation/pull/143. Just move the users creation to the finish client, so 2nd stage is not needed to have a minimal system.


src/clients/users_auto.rb
```diff
@@ -129,13 +129,8 @@ def main
...
       elsif @func == "Write"
-        Yast.import "Progress"
-        Users.SetWriteOnly(true)
-        @progress_orig = Progress.set(false)
-        @ret = Users.Write == ""
-        Progress.set(@progress_orig)
-      # Return if configuration  was changed
-      # return boolean
+        # NOTE: this code was moved to users_finish client, as users set up is
+        # done now in the 1st stage.
```
kobliha: `@ret` must be set to `true`

kobliha:
@imobach Looking at the change, you should IMO also add a conflict with `AutoYast < 3.1.92`

src/clients/users_auto.rb
imobach: Must we return false to make clear that this client is not going to write anything? (It shouldn't be called anyway with `write` action).

kobliha: Let's ask @schubi2 : IMO if we have a function, it has to return something, if it's never called, we should not have it either

imobach: @kobliha Yes, that's make sense. I think we should remove the `Write` action then. But @schubi2 will say :)

imobach:
@kobliha I agree. I've added the conflict to the spec file.

src/clients/users_auto.rb
schubi2: Hm, I would be carefully here because we have a use case in "yast2 autoyast" module where you can change settings and "apply" to the system via the write calls. Another use case is "yast ayast_setup" which is used in a running system. This functionality is mainly used by the SAP team. But I am not sure if they are using the user module too.

src/desktop/users.desktop
```diff
@@ -9,7 +9,7 @@ X-SuSE-YaST-Call=users
 X-SuSE-YaST-Group=Security
 X-SuSE-YaST-Argument=
 X-SuSE-YaST-RootOnly=true
-X-SuSE-YaST-AutoInst=all
+X-SuSE-YaST-AutoInst=configure
```
imobach: Is this change safe?

schubi2: I think we should set it back to "all" because this module can still be used in an installed system for writing user settings.

src/clients/users_finish.rb
```diff
@@ -66,11 +66,20 @@ def main
...
         # Creating all users and their environment
 
-        # write the root password
-        UsersSimple.Write
+        if Mode.autoinst
+          # Write imported users (during autoupgrade no changes are done)
+          Users.SetWriteOnly(true)
+          @progress_orig = Progress.set(false)
+          @ret = Users.Write == ""
+          Users.SetWriteOnly(true)
```
schubi2: Just to be sure: Users.SetWriteOnly(true) called twice ? Perhaps a comment would be nice :-)

schubi2:
Just two little comments. From my point of view LGTM. Decide by your own to make these two changes or not and merge it :-)
Congratulations, that has been a really "tricky" task to solve !

src/clients/users_finish.rb
imobach: Uh, no, it's just duplicated. Thanks for noticing.

imobach:
@schubi2 Thanks for your help! It have been funny :)

src/desktop/users.desktop
imobach: Ok. And as the user keys are removed from the profile before the second stage, the users_auto will do anything. I've tried manually and it works.

imobach:
Ok, before merging and during integration tests, I found a funny issue: if a package adds a new system user during installation, that user is clobbered during `users_finish`. The problem is that users are imported during `autosetup`, just before the installation.

As `Users` module does not have any method to `resync` data, I've decided to change `users_finish` a little bit to:

* Save imported data
* Read local users
* Import saved data

Integration tests are working again.

schubi2:
LGTM

PR #76, opened 2015-11-13
Update SLE 12 GA with IBS

imobach:
This PR merge changes up to 3.1.37 from `master` into `sle-12-ga`.


package/yast2-users.changes
```diff
@@ -2,10 +2,22 @@
 Mon Nov  9 10:52:39 UTC 2015 - igonzalezsosa@suse.com
 
 - fix validation of AutoYaST profiles (bsc#805275)
-- 3.1.35.1
+- 3.1.37.1
 
 -------------------------------------------------------------------
-Thu Nov 20 10:44:41 CET 2014 - jsuchome@suse.cz
```
mvidner: Here we'd be changing the timestamp of 3.1.35, which we must not do. Please keep the old timestamp.

It is a consequence of bad branching in the past, where a fix was made in #63 and #65 without merging, *and* with the same version for diverging branches.

imobach: I tried to keep in sync with the changes file which is already published in IBS: https://build.suse.de/package/view_file/SUSE:SLE-12:Update/yast2-users/yast2-users.changes?expand=1

Should I change it anyway?

kobliha: @mvidner See yast-devel, we have to change it

mvidner: Ah, I should have checked the IBS content, you are right.

mvidner:
LGTM.

imobach:
Thanks!

PR #81, opened 2015-12-17
Refactoring for `Make skipping user intuitive`

kobliha:
This is the maximum I can do now. The rest can be done in 2016 the earliest.
Tested manually.


src/modules/UsersUtils.rb
```diff
@@ -0,0 +1,51 @@
...
+# this program; if not, contact Novell, Inc.
+#
+# To contact Novell about this file by physical or electronic mail, you may find
+# current contact information at www.novell.com.
+# ------------------------------------------------------------------------------
+
+require "yast"
+
+module Yast
+  class UsersUtilsClass < Module
```
ancorgs: Extracting common parts sounds like a nice opportunity to start using objects. So instead of a good-for-anything `UsersUtils` module, what about a `UserPassword` class?

src/lib/users/clients/inst_root_first.rb
```diff
@@ -0,0 +1,274 @@
...
+
+      begin
+        return handle_ui_passwords
+      ensure
+        UsersSimple.UnLoadCracklib
+      end
+    end
+
+    # Validates whether password1 and password2 match and are valid
+    def validate_password(password_1, password_2)
```
ancorgs: See? Another candidate to be moved to the `UserPassword` class

ancorgs:
In general, there is a clear improvement. But I dislike the introduction of a new YCP-style module instead of going for objects. Change that and you got me :wink:

src/modules/UsersUtils.rb
kobliha: Good point.

src/lib/users/clients/inst_root_first.rb
kobliha: Sounds like a good idea, on the other hand, having `password_1` and `password_2` will be rather an exception in using this new class.

kobliha:
Thanks for your idea. Introducing new YCP-style module was the simplest idea that an old YaST hacker can have. Thinking beyond the old boundaries is the right way. I'll try to rework it when I have time for that.

src/clients/inst_root.rb
```diff
@@ -168,18 +162,8 @@ def main
...
-          Builtins.sformat(
-            # additional help text about password
-            _(
-              "<p>If you intend to use this password for creating certificates,\nit has to be at least %1 characters long.</p>"
-            ),
-            @pw_min_CA
-          )
-        )
-      end
+      validator = Users::CAPasswordValidator.new
```
jreidinger: Better use `::Users` otherwise you risk collision

src/clients/inst_root.rb
```diff
@@ -168,18 +162,8 @@ def main
...
+      @helptext << validator.help_text if validator.enabled?
```
jreidinger: what about tell, not ask and simply just use `@helptext << validator.help_text` and text_help method returns "" if not enabled?

src/lib/users/ca_password_validator.rb
```diff
@@ -0,0 +1,72 @@
...
+#  you may find current contact information at www.suse.com
+
+require "yast"
+
+module Users
+  # Validator to check if a password fulfills the requirements to
+  # generate CAs (see FATE#300438)
+  class CAPasswordValidator
+    MIN_LENGTH = 4
+    private_constant :MIN_LENGTH
```
jreidinger: nice, I do not know about that method :+1:

src/lib/users/ca_password_validator.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    Yast.import "ProductFeatures"
```
jreidinger: is there reason to have import inside class? it is a bit unintuitive for me as I expect it in require place and usually place it elsewhere only when there is benefit from it ( like rarely usage or circular dependencies )

src/lib/users/clients/inst_root_first.rb
```diff
@@ -0,0 +1,231 @@
...
+
+      # help text, continued 4
+      helptext << _(
+        "<p>\n" \
+        "Do not forget this \"root\" password.\n" \
+        "</p>"
+      )
+
+      validator = Users::CAPasswordValidator.new
+      helptext << validator.help_text if validator.enabled?
```
jreidinger: again same duplicite code ;)

src/lib/users/clients/inst_root_first.rb
```diff
@@ -0,0 +1,231 @@
...
+
+module Yast
+  # This library provides a simple dialog for setting new password for the
+  # system adminitrator (root) including checking quality of the password
+  # itself. The new password is not stored here, just set in UsersSimple module
+  # and stored later during inst_finish.
+  class InstRootFirstClient < Client
+    include Yast::Logger
+
+    def main
```
jreidinger: only method that have to be public, so rest can be mark as private.

src/lib/users/ca_password_validator.rb
ancorgs: Not really. I'll extract it.

src/lib/users/clients/inst_user_first.rb
```diff
@@ -0,0 +1,699 @@
...
+        message = passwd.errors.join("\n\n") + "\n\n" + _("Really use this password?")
+        if !Popup.YesNo(message)
+          return false
+        end
+      end
+
+      return true
+    end
+
+  end unless defined? InstUserFirstClient
```
jreidinger: no longer needed as you have it as lib and not client, so it won't be reevaluated unless some use `load` instead of `require`

src/lib/users/clients/inst_user_first.rb
```diff
@@ -0,0 +1,699 @@
...
+          UsersSimple.SetRootAlias(
+            UI.QueryWidget(Id(:root_mail), :Value) == true ? @username : ""
+          )
+        end
+      end
+
+      Wizard.CloseDialog if Mode.normal
+      Progress.set(@progress_orig)
+      @ret
+    end
```
jreidinger: This method hurts my eyes and heart, can it be work accident?

At least please separate some method to lazy loaded attributes in own method.

src/lib/users/clients/inst_user_first.rb
```diff
@@ -0,0 +1,699 @@
...
+      textdomain "users"
+
+      @text_mode = UI.TextMode
+
+      # full info about imported users
+      @imported_users = {}
+      # user names of imported users
+      @user_names = []
+      # names of imported users selected for writing
+      @to_import = []
```
jreidinger: many initialization can be in initialize method. In main for me should be only real actions like opening wizard.

src/lib/users/clients/inst_user_first.rb
```diff
@@ -0,0 +1,699 @@
...
+
+      # do not open package progress wizard window
+      @progress_orig = Progress.set(false)
+
+      @encryption_method = UsersSimple.EncryptionMethod
+
+      # radiobutton label
+      @button_label = _("L&ocal (/etc/passwd)")
+
+      @encoding2label = {
```
jreidinger: better use separate method for it

src/lib/users/clients/inst_user_first.rb
```diff
@@ -0,0 +1,699 @@
...
+
+        VBox(
+          Left(
+            RadioButton(
+              Id("users"),
+              Opt(:notify),
+              @button_label
+            )
+          ),
+          @import_available ?
```
jreidinger: better place it in own method. this ternary operator is pain.
something like
```
def import_checkbox
  return Empty unless @import_available

  HBox( ...
end
```

src/clients/inst_root.rb
ancorgs: I also considered it. I don't have a strong preference for any of the options but in the end this approach somehow felt more flexible and explicit to me. `help_text` does always what is expected (returning a help text) and is the dialog who decides if it's worth adding it to the big help text or not.

Intentionally adding an empty string to the big text help sounds a little bit more tricky. Doing something non-obvious for the sake of encapsulation.

src/lib/users/clients/inst_user_first.rb
```diff
@@ -0,0 +1,699 @@
...
+
+      if @users.size > 1
+        @to_import = @users.map { |u| u["uid"] || "" }
+      elsif @users.size == 1
+        if @users.first["__imported"] != nil
+          @to_import = [@users.first.fetch("uid", "")]
+        else
+          @user = @users.first
+        end
+      end
```
jreidinger: this deserve some comment as it looks very tricky

src/lib/users/clients/inst_user_first.rb
```diff
@@ -0,0 +1,699 @@
...
+
+      @user_type = @user.fetch("type", "local")
+      @username = @user.fetch("uid", "")
+      @full_name = @user.fetch("cn", "")
+      @password = @user["userPassword"]
```
jreidinger: this is a bit crazy. `@user` is set only if `@users.size == 1` and `@users.first["__imported"] == nil`, otherwise it is completely useless.

src/lib/users/clients/inst_user_first.rb
```diff
@@ -0,0 +1,699 @@
...
+
+      widgets.each do |w|
+        UI.ChangeWidget(Id(w), :Enabled, @to_import.empty?)
+      end
+
+      UI.SetFocus(Id(:full_name))
+
+      @login_modified = false
+
+      while true
```
jreidinger: `loop do`

src/lib/users/clients/inst_user_first.rb
```diff
@@ -0,0 +1,699 @@
...
+          _("<p>\nWith the current password encryption (%1), the password length should be between\n" +
+            " %2 and %3 characters.\n</p>"),
+          @encoding2label.fetch(@encryption_method, @encryption_method),
+          UsersSimple.GetMinPasswordLength("local"),
+          UsersSimple.GetMaxPasswordLength("local")
+        ) +
+        UsersSimple.ValidPasswordHelptext
+
+      validator = Users::CAPasswordValidator.new
+      help += validator.help_text if validator.enabled?
```
jreidinger: three times same code.

src/lib/users/clients/inst_user_first.rb
```diff
@@ -0,0 +1,699 @@
...
+        # help text for main add user dialog
+        _("<p>\nThe username and password created here are needed to log in " +
+          "and work with your Linux system. With <b>Automatic Login</b> enabled, " +
+          "the login procedure is skipped. This user is logged in automatically.</p>\n") +
+        # help text for main add user dialog
+        _( "<p>\nHave mail for root forwarded to this user by checking <b>Receive System Mail</b>.</p>\n")
+    end
+
+    # Helper function: ask user which users to import
+    def choose_to_import(all, selected)
```
jreidinger: two dialogs in one client? Does not help readability and reusability

src/lib/users/clients/inst_user_first.rb
```diff
@@ -0,0 +1,699 @@
...
+        if !Popup.YesNo(message)
+          return false
+        end
+      end
+
+      return true
+    end
+
+  end unless defined? InstUserFirstClient
+end
```
jreidinger: in general this class deserve a lot of love.

src/lib/users/local_password.rb
```diff
@@ -0,0 +1,119 @@
...
+
+module Users
+  # Password for a local user
+  #
+  # This class is used to validate the correctness of a password by checking
+  # some security requirements that the password must fulfill.
+  # The class only checks for non-fatal errors. If a password is considered
+  # invalid, it still can be used if the user decides to ignore the warnings.
+  class LocalPassword
+    Yast.import "UsersSimple"
```
jreidinger: same comment as before Yast.import is functionality equal require. Yast.include is functionality similar to ruby include. So I think we should have it in similar place and use it similarly

src/lib/users/local_password.rb
```diff
@@ -0,0 +1,119 @@
...
+    include Yast::I18n
+    include Yast::Logger
+
+    attr_reader :plain, :uid, :also_for_root
```
jreidinger: documentation?

src/lib/users/local_password.rb
```diff
@@ -0,0 +1,119 @@
...
+      @ca_validator ||= CAPasswordValidator.new
+    end
+
+  private
+
+    attr_accessor :validated
+    attr_writer :errors
+
+    def unset_validated
+      @errors = []
```
jreidinger: you define attr_writer and then do not use it?

src/lib/users/local_password.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+
+    def validate_strength
+      if !Yast::UsersSimple.LoadCracklib
+        log.error "loading cracklib failed, not used for pw check"
+        Yast::UsersSimple.UseCrackLib(false)
+      end
+
+      begin
+        @errors += call_check_password_ui
```
jreidinger: same here, writer not used

jreidinger: it is a bit confusing for me. Why not use opposite call with `with_cracklib` and below have 
`validate_strength` instead of `call_check_password_ui` which is for me quite bad name for method

src/lib/users/local_password.rb
```diff
@@ -0,0 +1,119 @@
...
+      }
+      if uid != "root"
+        args_for_cpui["root"] = also_for_root
+      end
+      Yast::UsersSimple.CheckPasswordUI(args_for_cpui)
+    end
+
+    def validate_ca_constraints
+      return unless ca_validator.enabled?
+      @errors += ca_validator.errors_for(plain)
```
jreidinger: again similar approach can be used and simply add nothing if not enabled.

jreidinger: For me it is similar like if you ask on meeting:
- Do you have something? -> yes -> What problems do you have?
or with my approach:
- What problems you have? nothing

src/lib/users/local_password.rb
ancorgs: It's a leftover. Good catch.

src/lib/users/local_password.rb
ancorgs: In fact, `ca_validator.errors_for` already returns the empty array if ca_validator is not enabled. The previous return is just an extra check (call me paranoid). I'll remove it, since it's confusing.

src/clients/inst_root.rb
jreidinger: It is probably just POV, as for me helptext provide info how to use functionality and if it is not enabled, then it return empty string, but maybe in future we decide to add something like `"Validating password for CA is not enabled. If you want to use it, do this and this"` so it is more like helptext for provided functionality. So not help for what is *can* provide, instead help for what it *really* provides.

src/lib/users/clients/inst_root_first.rb
ancorgs: done

src/lib/users/clients/inst_user_first.rb
ancorgs: Moved code. WONTFIX

src/lib/users/clients/inst_user_first.rb
ancorgs: Moved code. WONTFIX.

src/lib/users/clients/inst_user_first.rb
ancorgs: Moved code. WONTFIX.

src/lib/users/clients/inst_user_first.rb
ancorgs: Moved code. WONTFIX.

src/lib/users/clients/inst_user_first.rb
ancorgs: Moved code. WONTFIX. I would probably introduce a wrong comment (which is even worse)

src/lib/users/clients/inst_user_first.rb
ancorgs: Moved code. WONTFIX.

src/lib/users/clients/inst_user_first.rb
ancorgs: Moved code. WONTFIX.

src/lib/users/clients/inst_user_first.rb
ancorgs: Moved code. WONTFIX.

ancorgs:
@jreidinger Problems in new code (and also some of them in old code) addressed. WONTFIX for deep issues in old (just moved) code.

jreidinger:
LGTM, what about enabling coveralls when you add test helper support for it?

src/lib/users/clients/inst_user_first.rb
kobliha: In fact, there is some bug somewhere near here, but I did not report it yet and did not want to spend time on fixing what I'm not sure whether it's wrong. Human readable version: let do more refactoring when touching this piece of code next time, we will touch it anyway.

kobliha:
Thanks guys, I know that my refactoring was not perfect and that @ancorgs did what he could, but the intention of the refactoring was `to split it into functions and libraries for possibility to understand and test and for further refactoring while touching the code while implementing that feature`. It's just a preparation for next work. So, WIP, but already looks much better than before.

ancorgs:
@jreidinger test_helper already have simplecov (thus, coveralls) support. The problem is that there are only 5 ruby files in this project. 90% is Perl.

The two classes I added have 100% coverage. The rest of the code is not covered by RSpec. So simplecov is reporting a (completely fake) 26% of coverage, which is far from reality.

ancorgs:
@jreidinger Sorry, I misread the comment about coveralls and supporting it in test_helper. But the reality still stands, coveralls report will be unrealistic (too optimistic). Anyways, enabling it should be just a couple of clicks.

PR #82, opened 2016-01-12
Merge post SLE 12 SP1 release changes

imobach:
Merge SLE12 SP1 after release changes into SLE-12-SP1 to be released as a MU.


ancorgs:
LGTM. Travis does not agree :)

PR #84, opened 2016-01-25
Redesign and refactoring of installation dialogs

ancorgs:
This pull request tries to address the usability problems reported for the "create user" screen during installation.

## Problems of the previous dialog for user creation

Here you can see the previous screen.

createnewuser_-_old_state

There is quite some functionality hidden there:

* You can select the password encryption method (clicking on "Change...")
* You can select one and only one of these options:
  * Create a user (Obvious)
  * Skip user creation (Leaving the form blank and clicking "Next", see pop-up on screenshot)
  * Import users from a previous installation. (Really well hidden, clicking on "Change..." and then checking the corresponding checkbox -only visible if users are detected- under the section "authentication method"). Rather buggy sub-dialog.
* Last but not least, the dialog exposes a feature that is not there. The summary suggest that you can change the authentication method, but the only supported option is "local /etc/password".

Apart from the usability problems, the source code was in really bad shape. Poorly structured, with no unit tests, with a lot of duplication and dead code and with some bugs.

## The new dialog

The normal installation workflow have now 2 steps:

1) Creation of user. This is how it looks in a clean system and in a system with previous users
step1-clean
step1-import

2) Root's password. As usual, this step is skipped if "use this password for system administrator" is checked. It looks like always (not worth an additional screenshot).

That's it. No more hidden functionality.

But you could be wondering where the encryption method selector ended up. That depends on the content of the proposal section of the control.xml of the distribution.

**Option a)** Nothing (current) - The option is simply not there anymore

**Option b)** users proposal - The installation summary looks like this (check "users settings" section)
user_proposal

**Option c)** users_encryption proposal - Summary looks like this (check the "password encryption type" section)
encryption_proposal

With both options b and c, clicking on the name of the method will take the user to this screen
encryption

## Code refactoring

As important as the visual and workflow redesign it the code refactoring, which adds object orientation and some unit tests. As a side effect of that refactoring, the `InstallerDialog` class dialog was added to yast2.rpm for further reuse all along YaST in the future. See https://github.com/yast/yast-yast2/pull/427 for details


src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -0,0 +1,661 @@
...
+# $Id$
+require "yast"
+require "ui/event_dispatcher"
+require "users/dialogs/users_to_import"
+require "users/ca_password_validator"
+require "users/local_password"
+
+module Yast
+  class InstUserFirstDialog
+    include ::UI::EventDispatcher
```
jreidinger: why not use whole Dialog base class? https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/ui/dialog.rb

jreidinger: at least document why it cannot be used

src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -0,0 +1,661 @@
...
+
+      help
+    end
+
+  private
+
+    # Imports all the used YaST modules
+    #
+    # Importing them before class initialization has been observed to
+    # be problematic
```
jreidinger: problematic in which way?

src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -0,0 +1,661 @@
...
+
+    def initialize
+      import_yast_modules
+      textdomain "users"
+
+      @text_mode = UI.TextMode
+
+      # Widgets to enable/disable depending on the selected action
+      # (the first one receives the initial focus if applicable)
+      @widgets = {
```
jreidinger: looks like constant for me

src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -0,0 +1,661 @@
...
+        import: [:choose_users, :import_qty_label],
+        skip: []
+      }
+
+      # full info about imported users
+      @importable_users = {}
+      # user names of imported users
+      @importable_usernames = []
+      # names of imported users selected for writing
+      @usernames_to_import = []
```
jreidinger: looks like unnecessary as below is initialized - https://github.com/yast/yast-users/pull/84/files#diff-0998366812bd3d3dd504f9dcfaceedd3R100

src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -0,0 +1,661 @@
...
+      @users = UsersSimple.GetUsers
+      @action = case @users.size
+      when 0
+        # New user is the default option
+        GetInstArgs.going_back ? :skip : :new_user
+      when 1
+        @users.first["__imported"] ? :import : :new_user
+      else
+        :import
+      end
```
jreidinger: better to have it in own method, initializer is already quite big

src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -0,0 +1,661 @@
...
+    def import_handler
+      self.action = :import
+    end
+
+    # help text
+    def main_help
+      help = _(
+        "<p>\nUse one of the available options to add local users to the system.\n" \
+        "Local users are stored in <i>/etc/passwd</i> and <i>/etc/shadow</i>.\n</p>\n"
+      ) +
```
jreidinger: better use "\" instead of "+" as it is noop

src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -0,0 +1,661 @@
...
+
+    # Sets the initial default value for autologin
+    # Requires @user to be already set
+    def init_autologin
+      @autologin = UsersSimple.AutologinUsed
+      if @user.empty? && !@autologin
+        if ProductFeatures.GetBooleanFeature("globals", "enable_autologin") == true
+          @autologin = true
+        end
+        Builtins.y2debug("autologin default value: %1", @autologin)
```
jreidinger: better use new logging functionality

src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -0,0 +1,661 @@
...
+        "cn"           => @full_name
+      }
+      UsersSimple.SetUsers([user_map])
+      UsersSimple.SkipRootPasswordDialog(@use_pw_for_root)
+      if root_dialog_follows || @use_pw_for_root
+        UsersSimple.SetRootPassword(@use_pw_for_root ? @password : "")
+      end
+
+      UsersSimple.SetAutologinUser(
+        UI.QueryWidget(Id(:autologin), :Value) == true ? @username : ""
```
jreidinger: `== true` is not neccessary. It is ternary logic, and true is only valid that pass it

src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -0,0 +1,661 @@
...
+    end
+
+    def import_users
+      create_users = []
+      @usernames_to_import.each do |name|
+        u = @importable_users.fetch(name, {})
+        u["__imported"] = true
+        u["encrypted"] = true
+        create_users << u
+      end
```
jreidinger: looks like `map` is better then `each`

src/lib/users/dialogs/inst_user_first.rb
ancorgs: Dialog class assumes a non-wizard dialog (it calls to UI.CloseDialog, for example). Using it in this context means overriding almost the whole class (only the includes are really useful). We could, of course, also thing about modifying `Dialog` to place more nicely in these situations.

jreidinger: thanks for answer, so plase at least document it in code why it is not used. Dialog can be of course also adapted to work nicely in wizard dialogs

ancorgs: Expect a pull request today adding a `InstallationDialog` class to yast2 (I have already needed it twice).

jreidinger: Please do not call it InstallationDialog as wizard is used e.g. also in bootloader which is not only installation dialog. Better `WizardDialog`?

ancorgs: In its current form, it is really `InstallationDialog`, since it uses `Yast::GetInstArgs`, makes `accept_handler` kind of an alias for `next_handler` and that kind of installation specific stuff.

ancorgs: @jreidinger Here you are the premiere https://github.com/yast/yast-yast2/pull/427 I'll add tests tomorrow morning.

src/lib/users/dialogs/inst_user_first.rb
ancorgs: Removed from below

src/lib/users/dialogs/inst_user_first.rb
ancorgs: Reshuffled a little bit. Still, it's old code, so I consider it improved enough. :-)

src/lib/users/dialogs/inst_user_first.rb
ancorgs: This is not the only `y2debug` in this file. There is still quite some old code, but... you know.

src/clients/users_proposal.rb
```diff
@@ -76,134 +54,92 @@ def main
...
-        @ret = { "workflow_sequence" => @result }
+        ret = { "workflow_sequence" => result }
         Builtins.y2debug(
           "Returning from users_proposal AskUser() with: %1",
-          @ret
+          ret
         )
       end
-      deep_copy(@ret)
+      ret
```
jreidinger: if you touch this file I really recommend to use ProposalClient as base class - https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/installation/proposal_client.rb

src/lib/users/dialogs/encryption_method.rb
```diff
@@ -0,0 +1,78 @@
...
+      id = UI.QueryWidget(Id(:encryption_method), :CurrentButton)
+      method = ::Users::EncryptionMethod.new(id)
+      method.set_as_current
+      super
+    end
+
+    def help_text
+      help = _("<p>Choose a password encryption method for local and system users.</p>")
+      help += _(
+        "<p><b>%s</b> is the current standard hash method. Using other " \
```
jreidinger: please add help text for translators what is in %s e.g. if they can move it and other related stuff

src/lib/users/dialogs/encryption_method.rb
```diff
@@ -0,0 +1,78 @@
...
+        VStretch()
+      )
+    end
+
+    def method_widgets
+      widgets = ::Users::EncryptionMethod.all.map do |meth|
+        Left(RadioButton(Id(meth.id), meth.label, meth.current?))
+      end
+      # Let's add some vertical space after each widget
+      widgets.map {|w| [w, VSpacing(1)] }.flatten
```
jreidinger: using map and flatten smells for me a bit. btw even map can be replaced by `product([VSpacing(1)])`
what about using each_with_obj like:
```
::Users::EncryptionMethod.all.each_with_obj([]) do |meth, res|
   res << Left(RadioButton(Id(meth.id), meth.label, meth.current?))
   # Let's add some vertical space after each widget
   res << VSpacing(1)
end
```

src/lib/users/dialogs/users_to_import.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    def dialog_options
+      Opt(:decorated)
+    end
+
+    def items
+      @all.map {|u| Item(Id(u), u, @initial.include?(u)) }
+    end
+
+    def all_checked?
```
jreidinger: I found it a bit confusing as it looks like method that check if all checkboxes are checked, but it actually just check initial state. Maybe better name is enough like `initially_all_checked?`
same above with items, maybe use `initial_items` otherwise if someone modify it and use items to get fresh list, it do actually reset of selection.

src/lib/users/encryption_method.rb
```diff
@@ -0,0 +1,118 @@
...
+      # TRANSLATORS: encryption type
+      "sha512" => _("SHA-512")
+    }
+    private_constant :LABELS
+
+    DEFAULT_ID = "sha512"
+    private_constant :DEFAULT_ID
+
+    def initialize(id)
+      EncryptionMethod.raise_if_not_found(id)
```
jreidinger: I think `raise_if_not_found` do not looks like nice method name. What about `validate!(id)`

jreidinger: or check or something like this that shows that you want to verify id...say what you want to achieve and not what you want to have implemented.

src/lib/users/encryption_method.rb
```diff
@@ -0,0 +1,118 @@
...
+    class Error < RuntimeError
+    end
+
+    class NotFoundError < Error
+    end
+
+    extend Yast::I18n
+    textdomain "users"
+
+    attr_accessor :id
```
jreidinger: JFYI it can cause problems in *ancient* ruby before Object#object_id was named Object#id

src/lib/users/encryption_method.rb
```diff
@@ -0,0 +1,118 @@
...
+    def ==(other)
+      other.is_a?(EncryptionMethod) && other.id == id
+    end
+    alias_method :eql?, :==
+
+    # Internationalized name of the method
+    #
+    # @return [String]
+    def label
+      LABELS[id]
```
jreidinger: this doesn't work. Do not forget that yast allow change on language on fly, but this translations are searched only once.
Solutions are basically two
1) usage of N_ and _ ( like in https://github.com/yast/yast-packager/blob/master/src/modules/SourceDialogs.rb#L31 )
2) change constant LABELS to method

src/lib/users/encryption_method.rb
```diff
@@ -0,0 +1,118 @@
...
+    # List of all supported encryption methods
+    #
+    # @return [Array<EncryptionMethod>]
+    def self.all
+      LABELS.keys.sort.map { |id| new(id) }
+    end
+
+    # Current encryption method
+    #
+    # Raises an exception if the id returned by UsersSimple is unknown
```
jreidinger: yardoc have tag for it `@raise`

src/lib/users/encryption_proposal.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    def description
+      {
+        "id"              => "users_encryption",
+        # TRANSLATORS: rich text label
+        "rich_text_title" => _("Password Encryption Type"),
+        "menu_titles"     => [
+          # TRANSLATORS: menu button label
+          { "id" => "users--encryption", "title" => _("&Encryption Method") }
```
jreidinger: third time same id, constant?

jreidinger:
hidden value in else is medium serious for me. another one is nitpick. so almost LGTM :wink:

src/lib/users/dialogs/inst_user_first.rb
ancorgs: Quite random exceptions raised in `import`. Usually (but not always) while trying to import `Report`.

kobliha:
I have one little nitpicking: The dialog title is "Local Users" even if there is nothing to import and thus you can create just one "Local User". So, either I'd like to have it different dependent on this fact or just "Local User" (one).

jreidinger:
for me LGTM

kobliha:
LGTM, thanks for the changes!

PR #85, opened 2016-01-29
Small visual improvements in InstUserFirstDialog

ancorgs:


src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -496,10 +496,11 @@ def new_user_option
           RadioButton(
             Id(:new_user),
             Opt(:notify),
-            "Create new user",
+            "Create New User",
```
kobliha: As this is a clickable widget, we should usually assign a good shortcut for that `&`

BTW, this deserves a comment for translators - radio button

kobliha:
Some `NiceToHave` comments, then LGTM

kobliha:
Still LGTM :) Thank you!

PR #86, opened 2016-02-04
Remove system mail option

jreidinger:


package/yast2-users.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Feb  4 10:05:57 UTC 2016 - jreidinger@suse.com
+
+- drop receive system mail in installation (FATE#320448)
```
mvidner: > Drop the Receive System Mail checkbox from the installation workflow (FATE#320448)

jreidinger: thanks, will update it

mvidner:
Just enhance the changelog note, otherwise LGTM.

ancorgs:
If the fate is only about the dialog. Then it's fine. But have you checked if `GetRootAlias` and `SetRootAlias` are also used in autoyast or have a default value set by control.xml or something like that?  I think the plan was to also remove all the dead code about that feature.

jreidinger:
related pr  https://github.com/yast/yast-storage/pull/187 which removes one of removed method from other module where it doesn't make sense.

ancorgs:
Shouldn't we do something to ensure that this not get installed alongside an old version of yast2-storage (one requiring `UsersSimple.AfterAuth`to be there)?

ancorgs:
Nice to see so much code from `UsersSimple` being dropped. Do we have some mechanism to announce these API changes? I wouldn't be unlikely to have a third party module (let's say SAP stuff) using it. We should warn them somehow.

jreidinger:
@ancorgs We can add conflict. so I do it.

regarding API changes we do not have such mechanism but as UsersSimple is installation only ( for runtime usage there is Users module ), I do not expect much problems.

ancorgs:
LGTM

About the deprecation mechanism (well, "dropping" is a better word than "deprecation"), I will start a discussion in yast-devel

kobliha:
:clap:

PR #87, opened 2016-02-18
[Review] Request from 'schubi2' @ 'yast/yast-users/review_160218_checking_empty_string'

schubi2:
  * 5361e67 packaging
  * c96224b checking for empty string

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,7 +1,8 @@
 -------------------------------------------------------------------
 Thu Feb 18 09:23:10 CET 2016 - schubi@suse.de
 
-- Checking empty string while generating home directory.
+- Do not crash autoinstallation if the home directory of a user is
+  directly under /.
   (bnc#966867)
 - 3.1.45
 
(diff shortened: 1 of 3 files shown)


mvidner:
This is interesting!
I tested the situation (.target.stat of an empty string) and could not get a crash. Further inspection revealed that it does [`abort` in  SCRAgent::targetPath](https://github.com/yast/yast-core/blob/2f40b9caf6f7b8ada6293161b42b8c17c7425992/libscr/src/SCRAgent.cc#L125) if SCR has a target chroot set.

Cc: @jreidinger , we may have a similar problem in other cases.

package/yast2-users.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Feb 18 09:23:10 CET 2016 - schubi@suse.de
+
+- Checking empty string while generating home directory.
```
mvidner: No, here you're describing the internals which is not helpful to the user.
Better:
> Do not crash autoinstallation if home directory is directly under /.

mvidner:
LGTM

mvidner:
Bug filed: [ SCR crashes (abort) on an empty path inside chroot](https://bugzilla.opensuse.org/show_bug.cgi?id=967183)

PR #88, opened 2016-02-18
backport from master

schubi2:
This is a backport from master

### DIFF
--- a/package/yast2-users.spec
+++ b/package/yast2-users.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-users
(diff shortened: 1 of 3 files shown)


package/yast2-users.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-users
-Version:        3.1.41
+Version:        3.1.42
```
mvidner: 3.1.41.1 because 3.1.42 is something different in `master`

mvidner:
LGTM

PR #89, opened 2016-03-05
Fix users import

imobach:
This patch fixes users import during autoinstallation. It's related to [bsc#965852](https://bugzilla.suse.com/show_bug.cgi?id=965852).

The original problem is that before doing the import, all local users (inst-sys users) were read, so inst-sys users where merged with imported ones. In this stage, we should ignore inst-sys users. That's not a problem during autoupgrade, as long as users are not modified/created in that scenario.

### DIFF
--- a/src/modules/Users.pm
+++ b/src/modules/Users.pm
@@ -6154,7 +6154,7 @@ sub Import {
     # remove cache entries (#50265)
     UsersCache->ResetCache ();
 
-    my $error_msg = Mode->test () ? "" : $self->ReadLocal ();
+    my $error_msg = (Mode->test() || Stage->initial()) ? "" : $self->ReadLocal ();
     if ($error_msg) {
 	return 0;
     }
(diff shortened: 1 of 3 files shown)


src/modules/Users.pm
```diff
@@ -6154,7 +6154,7 @@ sub Import {
     # remove cache entries (#50265)
     UsersCache->ResetCache ();
 
-    my $error_msg = Mode->test () ? "" : $self->ReadLocal ();
+    my $error_msg = (Mode->test() || Stage->initial()) ? "" : $self->ReadLocal ();
```
schubi2: I would put a comment here. Perhaps the bugzilla nr. too.

schubi2:
Please add a comment. Relly nice and short fix for that complex stuff. --> LGTM

imobach:
Thanks for reviewing. You're right, I'll add that information as it's quite tricky.

schubi2:
Thanks ! LGTM

imobach:
@schubi2 Well, it turned out to be more complicated. The fix won't work if `<image_installation>` is set to `false`. So I've added a second fix to match users by username during 1st stage. So we're modifying the behavior only during 1st stage. Does sound it reasonable to you?

schubi2:
LGTM again :-)

imobach:
Thanks again :)

PR #91, opened 2016-04-07
Fix users import during installation

imobach:
Fixes bsc#973639 and bsc#971804.


src/clients/users_finish.rb
```diff
@@ -71,9 +73,12 @@ def main
 
           # During installation, some package could add a new user, so we
           # need to read them again before writing.
```
kobliha: Maybe the comment does not belong here, but down ... above the `Read()`?

src/clients/users_finish.rb
```diff
@@ -71,9 +73,12 @@ def main
...
-          Users.SetExportAll(true)
+          Users.SetExportAll(false) # Export only previously imported users
```
kobliha: And then you could move the comment just above?

src/clients/users_finish.rb
imobach: What I wanted to express is that we need to Export -> Re-read -> Import again (that's the reason to put the comment before the whole block).

But I see that's not clear at all, so I've added some extra documentation.

schubi2:
Uhi, you have fixed a lot of bnc#* !  :-) LGTM

src/clients/users_finish.rb
```diff
@@ -71,12 +73,20 @@ def main
...
+          Users.SetExportAll(false)
           saved = Users.Export
-          Users.ReadLocal
+          log.info("Users to import: #{saved}")
+
+          # 2. Read users and settings from the installed system
+          # (bsc#965852, bsc#973639, bsc#974220 and bsc#971804)
+          Users.Read
+
+          # 3. Import users exported in step 1
```
kobliha: It's actually `Merge users from the system with new users from AutoYast profile` ?

imobach:
I've moved `UsersFinish` client to `lib/users/clients`. BTW, I've done some refactoring and now it inherits from `::Installation::FinishClient`.

Adding tests was a little bit tricky because the code, as it's right now, is not easy to test. But I hope it's better now :)

kobliha:
LGTM (but I haven't checked whether it all works "as is", Travis should say his word here)

mvidner:
The matching integration test is https://github.com/yast/aytests-tests/pull/19.

package/yast2-users.spec
```diff
@@ -37,6 +37,7 @@ BuildRequires:  yast2-ldap >= 3.1.2
 BuildRequires:  yast2-perl-bindings
 BuildRequires:  yast2-security
 BuildRequires:  yast2-testsuite
+BuildRequires:  rubygem(rspec)
```
mvidner: `BuildRequires:  rubygem(%rb_default_ruby_abi:rspec)`

mvidner:
Thanks for the client refactoring and tests!
Just one packaging improvement,
otherwise LGTM.

imobach:
@mvidner Thanks! I've updated the spec file with your suggestion.

mvidner:
LGTM

imobach:
Thanks both for reviews. Merging after getting positive feedback.

PR #95, opened 2016-04-13
Fix /etc/shadow after installing yast2-users package

imobach:


package/yast2-users.spec
```diff
@@ -98,6 +98,29 @@ provided by yast2-users package.
...
+# warndays in /etc/shadow for a given set of users.
+# bin::16903:0:99999:7::: -> bin:*:16903::::::
+function set_password {
+  declare -a users=("${!2}")
+  regexp=$(IFS="|"; echo "${users[*]}")
+  sed /etc/shadow -i -re \
+    "s/^($regexp)::([^:]*):[^:]*:[^:]*:[^:]*:(.+)/\1:$1:\2::::\3/"
+}
+# Backup /etc/shadow
+cp /etc/shadow{,.YaST2save}
```
lslezak: I'd prefer `cp -a` to copy all file attributes.

kobliha:
Just for a reference, we have some default values in /var/adm/fillup-templates/shadow.aaa_base

PR #97, opened 2016-05-05
Fix users import when an entry in /etc/shadow is missing

imobach:
* Fixes [bsc#978648](https://bugzilla.suse.com/show_bug.cgi?id=978648).

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,8 +1,8 @@
 -------------------------------------------------------------------
 Thu May  5 12:32:39 UTC 2016 - igonzalezsosa@suse.com
 
-- Fix users import when an entry in /etc/shadow is missing
-  (bsc#978648)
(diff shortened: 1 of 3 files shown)


package/yast2-users.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu May  5 12:32:39 UTC 2016 - igonzalezsosa@suse.com
+
+- Fix users import when an entry in /etc/shadow is missing
```
ancorgs: NP: In fact it's when ALL entries for local users are missing.

ancorgs:
LGTM

package/yast2-users.changes
imobach: I've updated the message but using `non-system` instead of `local` (although we use the term `local` in the code) :)

imobach:
Thanks!

PR #98, opened 2016-05-05
Make users import independent of "copy_to_system"

ancorgs:
See comment on https://fate.suse.com/319624

This goes together with https://github.com/yast/yast-installation/pull/371


teclator:
LGTM

PR #101, opened 2016-06-01
Remove devel-doc package (fate#320356) & version bump

cwh42:


agent-crack/doc/crack.html
```diff
@@ -1,71 +0,0 @@
-<HTML>
```
jreidinger: What is reason to remove this document? Even if we do not distribute it, I think it is useful as source of information about SCR agent on top of libcrack

imobach:
Replaced by #102.

PR #102, opened 2016-06-02
Drop yast2-users-devel-doc package

imobach:
Drop yast2-users-devel-doc package ([fate#320356](https://fate.suse.com/320356)). Replaces #101.


package/yast2-users.spec
```diff
@@ -128,8 +119,4 @@ provided by yast2-users package.
 %doc %{yast_docdir}/TODO.txt
```
jreidinger: To be honest I am not sure if it is good idea to distribute TODO documentaiton.

jreidinger:
just NP, LGTM

imobach:
@jreidinger I agree. I've removed that documentation.

jreidinger:
LGTM

imobach:
Thanks!

PR #109, opened 2016-08-30
[Review] Request from 'schubi2' @ 'yast/yast-users/review_160830_checking_user_entries_in_autoyast'

schubi2:
  * 524cc9a packaging
  * a1a7a69 checking defined users in AutoYaST profile

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Tue Aug 30 10:37:18 CEST 2016 - schubi@suse.de
 
-- AutoYaST: Checking users entries for double username/UIDs.
+- AutoYaST: Checking users entries for duplicate username/UIDs.
   (bnc#995397)
 - 3.1.56
 
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Aug 30 10:37:18 CEST 2016 - schubi@suse.de
+
+- AutoYaST: Checking users entries for double username/UIDs.
+  (bnc#995397)
+- 3.1.56
+
+-------------------------------------------------------------------
 Fri Aug 12 14:37:40 UTC 2016 - ancor@suse.com
 
 - The YaST client using the user creation dialog can now know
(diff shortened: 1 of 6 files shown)


src/clients/users_auto.rb
```diff
@@ -36,8 +36,24 @@
...
+
 module Yast
   class UsersAutoClient < Client
+
+    # Checking double user entries
+    # (double username or UID)
+    # @param [Array] users to check
+    def check_users(users)
+      if users.size > users.uniq { |u| u["username"]}.size
+        Report.Error(_("Found users in profile with equal <usersname>."))
```
imobachgs: Typo? `<username>`

package/yast2-users.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Aug 30 10:37:18 CEST 2016 - schubi@suse.de
+
+- AutoYaST: Checking users entries for double username/UIDs.
```
imobachgs: I would say `duplicated` instead of `double`.

src/clients/users_auto.rb
schubi2: Yes, as usual :-)

test/users_auto_test.rb
```diff
@@ -0,0 +1,41 @@
...
+        let(:users) { YAML.load_file(FIXTURES_PATH.join("users_error.yml")) }
+
+        before do
+          allow(Yast::WFM).to receive(:Args).with(no_args).and_return([func,users])
+          allow(Yast::WFM).to receive(:Args).with(0).and_return(func)
+          allow(Yast::WFM).to receive(:Args).with(1).and_return(users)
+        end
+
+        it "report error" do
+          expect(Yast::Report).to receive(:Error).with("Found users in profile with equal <username>.")
```
imobachgs: This test may fail in systems with a different locale. You can use the `_` method here if you add these lines to `test_helper.rb`:

```ruby
RSpec.configure do |config|
  config.extend Yast::I18n  # available in context/describe
  config.include Yast::I18n # available in it/let/before/...
end
```

See https://github.com/yast/yast-packager/blob/master/test/test_helper.rb#L17 for an example.

src/clients/users_auto.rb
```diff
@@ -36,8 +36,24 @@
...
+
+Yast.import "Report"
+
 module Yast
   class UsersAutoClient < Client
+
+    # Checking double user entries
+    # (double username or UID)
+    # @param [Array] users to check
+    def check_users(users)
```
imobachgs: If that method is not part of the public API (and it's not), I would move it to the end and set it as `private`.

imobachgs:
LGTM if Travis is happy.

PR #110, opened 2016-09-02
Bug 996823 sp2

schubi2:


package/yast2-users.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Sep  2 15:16:37 CEST 2016 - schubi@suse.de
+
+- AuotYaST: Ignore Users with not UID while checking for duplicate
```
jreidinger: typo autoyast

jreidinger: and maybe "without" instead "with not"

src/clients/users_auto.rb
```diff
@@ -171,6 +171,8 @@ def check_users(users)
       if users.size > users.uniq { |u| u["username"]}.size
         Report.Error(_("Found users in profile with equal <username>."))
       end
+      # Do not check users without defined UID. (bnc#996823)
+      users.reject! { |u| !u.key?("uid")}
```
jreidinger: it modifies input parameter, is it OK? And if so, then please document it in method definition, but such nasty side effect should be avoided.

jreidinger:
LGTM

PR #112, opened 2016-09-15
Add authorized keys support to AutoYaST

imobachgs:
This PR adds support to specify SSH authorized keys in AutoYaST profiles. It's not finished at all but I need some feedback.

## Format

The user can specify the authorized keys like this:

```xml
<user>
  <username>suse</username>
  <authorized_keys config:type="list">
    <authorized_key>ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCpZC8ctjmn90B/MxLOdSjYM3Yl3qd+BhTWYdBNgO3B1fJ1JSegTgCpDM0krMHqd/OAslW5H3MRED7g7g9WkKZh5xTMGvH56yRitJySfSiK8uSxCu6Jg7NM11kqOs5/RwycHO8955QrEYyiWOx80unD+CBJxGEZCOu/DH3ca4yEigAt2HSuC8NPicmRJWua6IbDa+VSICvdOTdFTM8izScSd5WBFH1ULz0bBfLnyi/pIiMjuHB69AN4gsUGYgKjzUsnufKli+DmzACgVWTdQ3Ukax/4/wgXFMr3KsDNpTbn7ZZOKzPpIXpzlP9AwbHQdym6J2NAPYV+DDY3Kcr/vql9 dummy@example.net</authorized_key>
  </authorized_keys>
</user>
```

Or even like this:

```xml
<user>
  <username>suse</username>
  <authorized_keys config:type="list">
    <authorized_key>
      <comment>dummy@example.net</comment>
      <content>AAAAB3NzaC1yc2EAAAADAQABAAABAQCpZC8ctjmn90B/MxLOdSjYM3Yl3qd+BhTWYdBNgO3B1fJ1JSegTgCpDM0krMHqd/OAslW5H3MRED7g7g9WkKZh5xTMGvH56yRitJySfSiK8uSxCu6Jg7NM11kqOs5/RwycHO8955QrEYyiWOx80unD+CBJxGEZCOu/DH3ca4yEigAt2HSuC8NPicmRJWua6IbDa+VSICvdOTdFTM8izScSd5WBFH1ULz0bBfLnyi/pIiMjuHB69AN4gsUGYgKjzUsnufKli+DmzACgVWTdQ3Ukax/4/wgXFMr3KsDNpTbn7ZZOKzPpIXpzlP9AwbHQdym6J2NAPYV+DDY3Kcr/vql9</content>
      <keytype>ssh-rsa</keytype>
      <options>Tunnel="0"</options>
    </authorized_key>
  </authorized_keys>
</user>
```

To be honest, the second syntax looks better (as it's more expressive) but I understand that the first one is easier to use (just copy and paste). I'm also fine if we prefer to support only the first one (to be discussed).

## Implementation details

There're some implementation details the I would like to highlight here.

### Interacting with Perl code

Almost all `yast2-users` modules are written in Perl, so some interaction is needed. Currently, I've defined 4 classes:

* `SSHAuthorizedKeyring`: it's some kind of keyring which read/stores authorized keys for each home directory.
* `SSHAuthorizedKey`: represents an authorized key (with its four elements: options, keytype, content and comment).
* `SSHAuthorizedKeysFile`: represents `authorized_keys` files (read/write those files).
* `SSHAuthorizedKeys` (module): it implements a thin layer in to of the `SSHAuthorizedKeyring` so `Users.pm` can talk to the Ruby code. Also implements *import* and *export* functionality for each user (it's quite close to Perl code and I don't think it's reusable).

### SSH directory path
**updated: this code has been removed**

Currently `authorized_keys` lives in `$HOME/.ssh` directory. The code should work also if it's redefined in the future to something like `$HOME/.config/ssh` (it will take care of creating the directory structure if needed).

### Permissions

When writing a keys file, the owner/group of that file will be the same than the home directory (so no need to pass the user to the keyring). Permissions will be 0700 for directory and 0600 for `authorized_keys`.

## Should we simplify it?

I would like to hear from you if you want me to simplify this code a little bit:

* Getting rid of the second syntax in the profile. It would lead to simplify `SSHAuthorizedKey` (or even to remove it) as we'll only work with lines.
* ~~Removing support to multilevel SSH directories (I don't think it's likely to change in the future).~~ (removed)


jsrain:
Well, one syntax would be perfectly sufficient, if we have two different syntaxes, we need to keep them supported. OTOH, if it is implemented already, we should probably not remove them, there are reasons why both of them have their advantages.

kobliha:
In other words, the first option is a must have as it's the simplest implementation and users know how to deal with it. The other one makes the whole process more robust, understandable and explicit. It basically brings deeper understanding of what we are actually configuring and how. Additionally, the second option will be easier to document (key, value, description), while the first one is really just a copy&paste from a file. So, when we already have it implemented, and if we have an automatic CI test and unit tests and good documentation, they I'd be also for having both.

imobachgs:
@kobliha @jsrain Thanks for your comments. I've removed the code to handle multilevel `SSH` directory (less code to maintain and less code to break/fix) :)

Now we can start to review the code :)

src/lib/users/ssh_authorized_keyring.rb
```diff
@@ -0,0 +1,213 @@
...
+        file.save && set_owner_and_perms(path, owner, group, perms)
+      end
+
+      # Set owner and permissions for a given directory/file
+      #
+      # @param path  [String] Path to file/directory
+      # @param owner [Fixnum] Owner's UID
+      # @param group [Fixnum] Owner's GID
+      # @param perms [String] Permissions (in form "0700")
+      def set_owner_and_perms(path, owner, group, perms)
```
imobachgs: Maybe we could move this to some common place.

src/lib/users/ssh_authorized_key.rb
```diff
@@ -0,0 +1,156 @@
...
+    # * Content (mandatory): the key value itself.
+    # * Options (optional):  a command to be executed, environment variables,
+    #                        SSH configuration, etc.
+    # * Comment (optional):  useful to identify each key, although SSH will ignore it.
+    #
+    # @see man sshd(8)
+    class SSHAuthorizedKey
+      # Regular expression to parse `authorized_keys` lines. Taken
+      # (and modified) from:
+      # https://github.com/jordansissel/ruby-sshkeyauth/blob/master/lib/ssh/key/verifier.rb#L21
```
jsrain: NP: Will this URL remain persistant?

imobachgs: No, you're right, I'll update it.

src/lib/users/ssh_authorized_keyring.rb
jsrain: Actually, here I wonder why we don't use any native ruby functions and use SCR to manipulate file own/grp/mod

imobachgs: Because we need to take care of the `.target` thingie.

mvidner:
> To be honest, the second syntax looks better (as it's more expressive)

Do we need the expressivity? Before I look at the code, I think that we handle the SSH key as an opaque cookie, so we would only complicate our code needlessly.

mvidner:
@kobliha The Fate entry is in a kind of limbo as it is in a Rejected state for the current product version but it has no tags for the next product version.

mvidner:
Aww, so much code and texts and documentation is dealing with the individual parts of an authorized key. I really think we are overengineering this and should treat it as an opaque line of text.

I may be wrong. What do others think?

lslezak:
We do not offer any UI for configuring this feature, that means it is designed for advanced admins. And in that case I'd prefer the simpler (first) XML format.

Splitting the string into several parts does not add any value, e.g. you cannot change the key type without changing the key itself. 

And for admins it would make the configuration (writing the XML) much more difficult, you would always need to consult it with the doc how to correctly split the config into the parts.

Simple copy&paste from `id_rsa.pub` should be preferred, no additional "magic" which might be confusing for users. Use what they already know, do not force them to learn something new, esp. when there is no real reason why.

See e.g. https://github.com/settings/keys, there is a single input field for pasting the whole key, there are no separate fields for each value.

src/lib/users/ssh_authorized_keyring.rb
lslezak: Oh... then you could use [Yast::Execute.on_target](https://github.com/yast/yast-yast2/blob/master/library/system/src/lib/yast2/execute.rb#L41).

IMHO we should avoid using SCR unless really needed, personally I consider it as another "YCP zombie"...

imobachgs:
Ok, I've prepared a simplified version in PR #113. Thanks for your feedback!

imobachgs:
*See e.g. https://github.com/settings/keys, there is a single input field for pasting the whole key, there are no separate fields for each value.*

I see your point. Anyway, just to make this clear, it's not exactly the same case: despite its name, the `authorized_keys` does no only contains keys: you can specify options (command to execute when the user logs in, environment variables, etc.) and I'm not sure if you can such a thing in Github.

PR #113, opened 2016-09-16
Authorized keys simplified

imobachgs:
This is a simplified version of #112. Keys are reduced to text lines in the profile removing the `SSHAuthorizedKey` class.


src/lib/users/ssh_authorized_keyring.rb
```diff
@@ -0,0 +1,213 @@
...
+      # @return [Array<SSHAuthorizedKey>] List of authorized keys
+      def [](home)
+        keys[home] || []
+      end
+
+      # Determines if the keyring is empty
+      #
+      # @return [Boolean] +true+ if it's empty; +false+ otherwise
+      def empty?
+        keys.empty?
```
lslezak: NP: You could use a delegator to forward the method to the underlaying object (see e.g. https://github.com/yast/yast-registration/blob/master/src/lib/registration/addon.rb#L100).

But in this case for a single method a simple wrapper is preferred, just consider it when adding some more wrappers...

src/lib/users/ssh_authorized_keyring.rb
```diff
@@ -0,0 +1,213 @@
...
+      end
+
+      # Read keys from a given home directory and add them to the keyring
+      #
+      # @param path [String] User's home directory
+      # @return [Boolean] +true+ if some key was found
+      def read_keys(home)
+        path = authorized_keys_path(home)
+        return false unless FileUtils::Exists(path)
+        authorized_keys = Yast::Users::SSHAuthorizedKeysFile.new(path).keys
```
lslezak: Um, some methods use the `Yast::` prefix, some do not, please unify. (I'd prefer removing it except for `FileUtils` where I'd explicitly use it to avoid possible conflict with `::FileUtils`.

src/lib/users/ssh_authorized_keyring.rb
```diff
@@ -0,0 +1,213 @@
...
+        if !FileUtils::Exists(home)
+          log.error("Home directory '#{home}' does not exist!")
+          raise HomeDoesNotExist
+        end
+        owner, group = perms_from(home)
+        if create_ssh_dir(home, owner, group, SSH_DIR_PERMS)
+          write_file(home, owner, group, AUTHORIZED_KEYS_PERMS)
+        else
+          log.error("SSH directory does not exist and could not be created: giving up")
+          return false
```
lslezak: The error handling is a bit strange here, for missing home dir an exception is raised but when creating the SSH directory fails it returns `false`.

Either unify the behavior or document it in the method description (+ write the reason why).

src/lib/users/ssh_authorized_keyring.rb
```diff
@@ -0,0 +1,213 @@
...
+      # @return [Boolean] +true+ if keys were written; +false+ otherwise
+      def write_keys(home)
+        return false if keys[home].nil?
+        if !FileUtils::Exists(home)
+          log.error("Home directory '#{home}' does not exist!")
+          raise HomeDoesNotExist
+        end
+        owner, group = perms_from(home)
+        if create_ssh_dir(home, owner, group, SSH_DIR_PERMS)
+          write_file(home, owner, group, AUTHORIZED_KEYS_PERMS)
```
lslezak: IMHO passing the `SSH_DIR_PERMS` and `AUTHORIZED_KEYS_PERMS` is not need, the constants could be used directly in the called methods. Prefer less parameters, they are not called with different parameters elsewhere and they are private anyway...

imobachgs: Yes, I follow certain *functional mindset* for this kind of helpers method. I'll remove them.

src/lib/users/ssh_authorized_keyring.rb
imobachgs: Ok, I'll raise an exception in both cases as it a quite unexpected error.

src/lib/users/ssh_authorized_keyring.rb
```diff
@@ -0,0 +1,213 @@
...
+      # should support changing SSH_DIR to something like `.config/ssh`.
+      #
+      # @param home  [String] Home directory where SSH directory must be created
+      # @param owner [Fixnum] Owner's UID
+      # @param group [Fixnum] Owner's GID
+      # @param perms [String] Permissions (in form "0700")
+      # @return [String] Returns the path to the first created directory
+      def create_ssh_dir(home, owner, group, perms)
+        ssh_dir = ssh_dir_path(home)
+        return ssh_dir if FileUtils::Exists(ssh_dir)
```
lslezak: I'm thinking about the security issues in the code. When we are writing into the user's home we should consider it as a not trusted "input". E.g. users could make `.ssh` a symlink pointing to into another location possibly rewriting some files.

You could easily point to another user's home and override that SSH key and gain access to that account! :boom:  That's not a problem in AutoInstallation, but could be a problem in the AutoUpgrade mode...

I think here we should simply check for a symlink and abort (raise exception) in that case.

Additionaly it would be probably good to call `set_owner_and_perms(ssh_dir,...)` always to make sure the directory has correct permissions. (Or check them and abort if they are wrong.)

src/lib/users/ssh_authorized_keyring.rb
```diff
@@ -0,0 +1,213 @@
...
+      # @param path  [String] Path to file/directory
+      # @param owner [Fixnum] Owner's UID
+      # @param group [Fixnum] Owner's GID
+      # @param perms [String] Permissions (in form "0700")
+      def write_file(home, owner, group, perms)
+        path = authorized_keys_path(home)
+        file = SSHAuthorizedKeysFile.new(path)
+        file.keys = keys[home]
+        log.info "Writing #{keys[home].size} keys in #{path}"
+        file.save && set_owner_and_perms(path, owner, group, perms)
```
lslezak: This might be also a possible issue, the file is written and **then** the permissions are set. There is a theoretic chance that someone can read the data in the meantime. (Well, it depends on the default `umask`, but anyway...)

On the other hand we are writing the *public* SSH key so you cannot abuse the file even if you read it.

So try to fix that if possible, if not then at least document it.

src/lib/users/ssh_authorized_keyring.rb
imobachgs: AFAIK, users are not set up during autoupgrade (I've to check). Anyway, making the code more robust won't harm.

src/lib/users/ssh_authorized_keyring.rb
imobachgs: We could create an empty file with right owner/permissions and write it afterward. It should no be that hard. I'll give it a try.

src/lib/users/ssh_authorized_keyring.rb
lslezak: Yes please, make it robust. Even if it is not used now someone can use it in the future. Think also of some other cases (applying a profile in installed system or the firstboot mode).

src/lib/users/ssh_authorized_keyring.rb
lslezak: Check some other modules, I think this must be already solved somewhere else...

src/lib/users/ssh_authorized_keyring.rb
imobachgs: Yes, they're. I've to check what happens with existing users. For example, we don't want to overwrite its authorized keys, right?

On the other hand, I wouldn't never have figured out a scenario like that. Thanks :)

src/lib/users/ssh_authorized_keyring.rb
lslezak: `File.new` or `File.open` have optional `perm` argument. http://ruby-doc.org/core-2.2.1/File.html#method-c-open

lslezak: For the `SCR` API use the old&good `umask 0077  &&  touch foo` approach like [here](https://github.com/yast/yast-ca-management/blob/master/src/include/ca-management/crlExport.rb#L281).

src/lib/users/ssh_authorized_keyring.rb
lslezak: Owerwriting is OK IMHO (if applying in installed system), the problem is that we need to check we are really overwriting the SSH key, not something else pointed by a malicious symlink...

imobachgs:
I've updated the PR with some changes:

* Now it uses `FileUtils->Chown` and `FileUtils->Chmod` functions (I overlooked them).
* Add errors handling.
* `authorized_keys` are written during autoinstallation and autoupgrade.
* `authorized_keys` is created with the correct permissions from the beginning (thanks @lslezak for the umask trick).

src/lib/users/ssh_authorized_keyring.rb
```diff
@@ -139,71 +142,34 @@ def authorized_keys_path(home)
...
+      # @param user  [Fixnum] Users's UID
+      # @param group [Fixnum] Group's GID
       # @return [String] Returns the path to the first created directory
-      def create_ssh_dir(home, owner, group)
+      #
+      # @raise SSHDirectoryIsLink
+      # @raise CouldNotCreateSSHDirectory
+      def create_ssh_dir(home, user, group)
         ssh_dir = ssh_dir_path(home)
+        raise SSHDirectoryIsLink if FileUtils::IsLink(ssh_dir)
```
lslezak: NP: Maybe we should invert the condition: `raise ... if FileUtils::Exists(ssh_dir) && !FileUtils::IsDir(ssh_dir)` to handle also other non-dir cases.

src/modules/SSHAuthorizedKeys.rb
```diff
@@ -39,6 +45,27 @@ def read_keys(home)
     # @see Yast::Users::SSHAuthorizedKeyring#write_keys
     def write_keys(home)
       keyring.write_keys(home)
+    rescue Users::SSHAuthorizedKeyring::HomeDoesNotExist
+      log.warn("Home directory '#{home}' does not exist")
+      Report.Warning(
+        # TRANSLATORS: '%s' is a directory path
+        _(format("Home directory '%s' does not exist\n" \
```
lslezak: It's the other way round, first translate then format. Otherwise the translation won't be found. I.e. use `format(_("..."), foo)`.

src/modules/SSHAuthorizedKeys.rb
```diff
@@ -39,6 +45,27 @@ def read_keys(home)
...
+                 "so authorized keys won't be written.", home))
+      )
+    rescue Users::SSHAuthorizedKeyring::SSHDirectoryIsLink
+      log.warn("SSH directory under '%s' is a symbolic link.")
+      Report.Warning(
+        # TRANSLATORS: '%s' is a directory path
+        _(format("SSH directory under '%s' is a symbolic link.\n" \
```
lslezak: Same here...

src/modules/SSHAuthorizedKeys.rb
```diff
@@ -39,6 +45,27 @@ def read_keys(home)
...
+                 "It may cause a security issue so authorized\n" \
+                 "keys won't be written.", home))
+      )
+    rescue Users::SSHAuthorizedKeyring::CouldNotCreateSSHDirectory
+      log.warn("SSH directory under '#{home}' could not be created")
+      Report.Warning(
+        # TRANSLATORS: '%s' is a directory path
+        _(format("Could not create SSH directory under '%s',\nso authorized keys won't be written.", home))
```
lslezak: ... and here.

src/modules/SSHAuthorizedKeys.rb
```diff
@@ -39,6 +45,27 @@ def read_keys(home)
     # @see Yast::Users::SSHAuthorizedKeyring#write_keys
     def write_keys(home)
       keyring.write_keys(home)
+    rescue Users::SSHAuthorizedKeyring::HomeDoesNotExist
+      log.warn("Home directory '#{home}' does not exist")
+      Report.Warning(
+        # TRANSLATORS: '%s' is a directory path
+        _(format("Home directory '%s' does not exist\n" \
+                 "so authorized keys won't be written.", home))
```
lslezak: IIRC abbreviations are used in informal language, in computer UI we should avoid that.

src/modules/Users.pm
```diff
@@ -4561,6 +4571,9 @@ sub Write {
 	    }
 	}
     }
+
+    if (Mode->autoinst() || Mode->autoupgrade() || Mode->config()) { WriteAuthorizedKeys(); }
```
lslezak: You can use `Mode->auto()` which covers both [`autoinst` and `autoupgrade`](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Mode.rb#L289) modes.

test/lib/users/ssh_authorized_keyring_test.rb
```diff
@@ -167,7 +161,16 @@
         end
       end
 
-      context "when ssh directory already exists" do
+      context "when SSH directory is a symbolic link" do
+        it "raises a SSHDirectoryIsLink" do
```
lslezak: I'd make the assertion more strong with adding `... and does not write the SSH key` + add expectation `#save` to not be called.

test/ssh_authorized_keys_test.rb
```diff
@@ -29,69 +31,78 @@
...
-  before do
-    allow(Yast::Users::SSHAuthorizedKey).to receive(:build_from)
-      .with(valid_key_spec)
-      .and_return(key)
-    allow(Yast::Users::SSHAuthorizedKey).to receive(:build_from)
-      .with(invalid_key_spec)
-      .and_return(nil)
-    ssh_authorized_keys.main
-  end
+  before { subject.main }
```
lslezak: Calling `main` explicitly is probably not needed anymore, it has been added [here](https://github.com/yast/yast-users/pull/113/files/4eca84cc0740f7cdbfd8ce1e8f80bd0e5905ed45..3d8a669442be7097662b58abdeab41cad0f4f6b8#diff-f2c8454f72a2b2ea8ee93bc0ab18fce1R97).

test/ssh_authorized_keys_test.rb
```diff
@@ -29,69 +31,78 @@
...
+    end
 
-      it "registers given keys" do
-        subject.import_keys(home, keys)
-        expect(subject.keyring[home]).to eq([key])
-      end
+    it "returns false if no key was imported" do
+      allow(subject.keyring).to receive(:add_keys).with(home, keys)
+        .and_return(keys)
+      expect(subject.import_keys(home, keys)).to eq(true)
```
lslezak: Um, the `it` title says `returns false`... :question: :question: :question:

test/ssh_authorized_keys_test.rb
```diff
@@ -29,69 +31,78 @@
...
-      let(:key) do
-        double("key", options: "tunnel=0", keytype: "ssh-rsa", content: "123ABC",
-          comment: "user@example.net")
+  describe "#write_keys" do
+    context "when home directory does not exists" do
+      it "shows an error message" do
+        allow(subject.keyring).to receive(:write_keys)
+          .and_raise(Yast::Users::SSHAuthorizedKeyring::HomeDoesNotExist)
+        expect(Yast::Report).to receive(:Warning)
+          .with("Home directory '#{home}' does not exist\n" \
```
lslezak: I'd be less strict about testing exact error messages, e.g. something like `with(/'#{home}' does not exist/)` should be enough. Shorter condition and smaller chance to break a test after fixing the message.

More over it does not add any value, you probably did copy&paste from the code. If there is a typo or a grammar issue it will be here as well.

test/ssh_authorized_keys_test.rb
```diff
@@ -29,69 +31,78 @@
...
-          "comment" => key.comment,
-          "content" => key.content,
-          "keytype" => key.keytype
-        }])
+    context "SSH directory is a link" do
+      it "shows an error message" do
+        allow(subject.keyring).to receive(:write_keys)
+          .and_raise(Yast::Users::SSHAuthorizedKeyring::SSHDirectoryIsLink)
+        expect(Yast::Report).to receive(:Warning)
+          .with("SSH directory under '#{home}' is a symbolic link.\n" \
```
lslezak: Same here...

test/ssh_authorized_keys_test.rb
```diff
@@ -29,69 +31,78 @@
...
 
-    context "when no key was registered for the given home" do
-      it "returns an empty array" do
-        expect(subject.export_keys(home)).to eq([])
+    context "SSH directory could not be created" do
+      it "shows an error message" do
+        allow(subject.keyring).to receive(:write_keys)
+          .and_raise(Yast::Users::SSHAuthorizedKeyring::CouldNotCreateSSHDirectory)
+        expect(Yast::Report).to receive(:Warning)
+          .with("Could not create SSH directory under '#{home}',\n"\
```
lslezak: ...and here.

test/ssh_authorized_keys_test.rb
imobachgs: We need to reset the keyring, so it's needed.

test/ssh_authorized_keys_test.rb
imobachgs: Actually there's an error: the idea was to use the `_` method  to avoid i18n problems.

src/modules/Users.pm
imobachgs: @lslezak no, it' something I'll fix soon: `auto` is not published, so it's no accessible :(

test/ssh_authorized_keys_test.rb
lslezak: OK, then please add a comment why, this is quite unusual in YaST. Or add a `reset` method to be clear...

src/modules/Users.pm
lslezak: Ouch, I didn't know that issue...

test/ssh_authorized_keys_test.rb
lslezak: Actually `reset` should be preferred, `main` is a kind of module constructor and should be called only once when the module is loaded.

src/modules/SSHAuthorizedKeys.rb
imobachgs: Oh, I don't know what I was thinking about when I wrote that code...

src/modules/SSHAuthorizedKeys.rb
```diff
@@ -0,0 +1,104 @@
...
+        # TRANSLATORS: '%s' is a directory path
+        format(_("'%s' exists but it is not a directory. It may\n" \
+                 "a security issue so authorized keys will not\n" \
+                 "be written."), e.directory)
+      )
+    rescue Users::SSHAuthorizedKeyring::CouldNotCreateSSHDirectory => e
+      log.warn(e.message)
+      Report.Warning(
+        # TRANSLATORS: '%s' is a directory path
+        format(_("Could not create SSH directory under '%s',\n" \
```
lslezak: NP: Maybe instead of *Could not create SSH directory under '/home/foo'* display something like *Cannot create directory '/home/foo/.ssh'*, that's IMHO more clear about what exactly failed.

And actually you should reuse the already existing [Message.UnableToCreateDirectory](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Message.rb#L234) method, no need for a new translation.

imobachgs: Indeed the message is wrong (as long as `e.directory` points to the `.ssh` one).

PR #114, opened 2016-09-20
Don't write authorized_keys if it's not  a regular file

imobachgs:
Fix a potential security issue in #113 (writing to authorized_keys when it's not a file).


package/yast2-users.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Sep 20 15:10:23 UTC 2016 - igonzalezsosa@suse.com
+
+- Prevent a potential security issue if authorized_keys file is
```
lslezak: NP: Just to make it clear: if **the target** "authorized_keys" file is not...

imobachgs: Thanks! fixed now.

imobachgs:
Thanks for the quick review.

PR #116, opened 2016-10-04
[Review] Request from 'schubi2' @ 'yast/yast-users/review_161004_checking_home_mount_points'

schubi2:
  * a824a08 checking all possible /home mount points

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Oct  4 13:12:40 CEST 2016 - schubi@suse.de
+
+- Checking all possilbe /home mount points (/mnt/home and /home).
+  (bnc#995299)
+- 3.1.61
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


package/yast2-users.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Oct  4 13:12:40 CEST 2016 - schubi@suse.de
+
+- Checking all possilbe /home mount points (/mnt/home and /home).
```
imobachgs: `possilbe` -> `possible`

imobachgs:
And BTW, what about Leap? Leap 42.2 will be based on SLE 12 SP2, so we need to make sure that this fix is available for Leap (but not sure about SLE).

schubi2:
The error happens only in Thumbl.  I have tested the fix with SLE12 and Leap too.

imobachgs:
Yeah, it was more about submitting it to the correct target.

imobachgs:
LGTM

PR #117, opened 2016-10-11
Fix the invocation of authentication configuration instance. 

HouzuoGuo:
Bump version to 3.1.62 for bsc#1000749.


ancorgs:
`master` is the right branch to fix this for tumbleweed and future releases (42.3, SP3) only. If we want to fix it for 42.2 and release it as an update for SP2 to another branch... that is not created. I will do it now.

PR #118, opened 2016-10-11
Fix version numbers to make maintenance possible

ancorgs:
I need to do this in order to create a consistent `merge_after_SP2_release` branch.

And please remember that changes introduced only in master should make numbers jump to 3.2.


imobachgs:
LGTM

PR #124, opened 2016-11-09
SLE 12 SP2 merge

imobachgs:
Merge `merge_after_SP2_release` branch into `SLE-12-SP2`.

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -3,7 +3,7 @@ Tue Oct 11 12:10:21 UTC 2016 - hguo@suse.com
 
 - Fix the invocation of authentication configuration instance
   (bsc#1000749).
-- 3.1.58
+- 3.1.57.1
 
 -------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


package/yast2-users.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Oct 11 12:10:21 UTC 2016 - hguo@suse.com
+
+- Fix the invocation of authentication configuration instance
+  (bsc#1000749).
+- 3.1.58
```
mvidner: Please renumber this to 3.1.57.1, because 3.1.58 is in 4be5ae3d872776304ce9a9cf2da65d551a54c9cb with different contents already.

imobachgs: Oh, I see. The problem is that 3.1.58 was later "renumbered" in e0f3db22.

imobachgs:
Thanks!

PR #126, opened 2016-12-01
Reusable root widget

jreidinger:
tested manually that it keep working in common installation

### DIFF
--- a/test/widgets_test.rb
+++ b/test/widgets_test.rb
@@ -1,0 +1,92 @@
+#!/usr/bin/env rspec
+
+require_relative "test_helper"
+require "users/widgets"
+
+def stub_widget_value(id, value)
+  allow(Yast::UI).to receive(:QueryWidget).with(Id(id), :Value).and_return(value)
+end
+
+describe Users::PasswordWidget do
+  it "has help text" do
+    expect(subject.help).to_not be_empty
+  end
+
+  it "has valid content" do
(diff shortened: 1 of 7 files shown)


test/widgets_test.rb
```diff
@@ -0,0 +1,92 @@
...
+      expect(Yast::UsersSimple).to receive(:CheckPassword).with("mimic_forbidden", "local").
+        and_return("Invalid password")
+      expect(Yast::Report).to receive(:Error)
+      expect(Yast::UI).to receive(:SetFocus).with(Id(:pw1))
+
+      expect(subject.validate).to eq false
+    end
+
+    it "asks for confirmation if password is not strong enough" do
+      stub_widget_value(:pw1, "a")
```
lslezak: NP: This test is a bit tricky, at the first sight it might seem that after setting the password to something like `eh3Roh5iefur2eew` the test should fail. But the validation is actually mocked, the value here is ignored.

I'd change the title to something some generic, like `asks for confirmation if password validation fails`.

jreidinger: ah, now I get it. I do not want to depend on cracklib as it can change its validation.

jreidinger: And I try to have it as unit as possible as dependency on cracklib agent is a bit tricky, as it have to be compiled and installed, which we currently do not do.

PR #128, opened 2016-12-02
Proper focus casp

jreidinger:


package/yast2-users.changes
```diff
@@ -1,9 +1,16 @@
 -------------------------------------------------------------------
+Fri Dec  2 12:58:58 UTC 2016 - jreidinger@suse.com
+
+- set focus for first password field to keep same behavior as
```
imobachgs: as -> than (not sure)

jreidinger: it is "as" ~ "like" @mvidner ?

mvidner: "the same as".
> Set the focus *to* *the* first password field to keep *the* same behavior as before *the* last change.

PR #131, opened 2016-12-15
Casp merge

jreidinger:


package/yast2-users.spec
```diff
@@ -30,7 +30,7 @@ BuildRequires:  libtool
 BuildRequires:  perl-Digest-SHA1
 BuildRequires:  perl-XML-Writer
 BuildRequires:  update-desktop-files
-BuildRequires:  yast2
+BuildRequires:  yast2 >= 3.2.8
```
imobachgs: Should we add the `UI::Widgets` comment here? I know it's written below but...

PR #137, opened 2017-02-28
Reduce perl

jreidinger:


src/modules/UsersDialogsFlags.rb
```diff
@@ -0,0 +1,30 @@
...
+      @start_dialog = name
+    end
+
+    # define dialog which should be used as first one, can be modified with {start_dialog=}
+    def start_dialog
+      # summary is starting dialog for installation
+      @start_dialog ||= "summary"
+    end
+
+    def assign_use_next_time(value)
```
lslezak: The usual prefixes are `get` and `set`, I'm fine with leaving out `get` but I'd rather prefer `set` instead of `assign`, e.g. use `set_use_next_time` here.

src/modules/Users.pm
```diff
@@ -3979,7 +3829,6 @@ sub WriteLoginDefaults {
 
 ##------------------------------------
 # Save Security settings (encryption method) if changed in Users module
-BEGIN { $TYPEINFO{WriteSecurity} = ["function", "boolean"]; }
```
lslezak: I'd keep the type data in a comment, it might be useful later when reimplementing in Ruby. Maybe you could use the yardoc format now so the refactoring it will be easier in the future...

lslezak:
BTW fix also the spec file, see the [Travis log](https://travis-ci.org/yast/yast-users/builds/206126172#L2027) for more details.

src/modules/Users.pm
jreidinger: ok, keeping in comment make sense.

teclator:
@jreidinger do you plan to work on it? There is still some changes pending. If we do not plan to work on it during a while maybe we could close it by now.

teclator:
@jreidinger closing it by now. Feel free to reopen it if you plan to finish the requested changes.

PR #138, opened 2017-03-07
Fix weak pwd2

jreidinger:


test/widgets_test.rb
```diff
@@ -84,6 +84,20 @@ def stub_widget_value(id, value)
       expect(subject.validate).to eq false
     end
 
+    it "asks for for confirmation only once for same password" do
```
teclator: hehe, now password is only asked once but we have for twice 😃

PR #144, opened 2017-08-07
Caasp merge

jreidinger:


package/yast2-users.changes
```diff
@@ -1,4 +1,44 @@
 -------------------------------------------------------------------
+Tue Mar  7 13:49:49 UTC 2017 - jreidinger@suse.com
+
+- do not ask again for already approved weak password second
+  attempt, so it will work also if widget is recreated(bsc#1025835)
+- 3.1.57.7
```
teclator: Should not it be 3.1.58?

jreidinger: previous one in this branch already have 3.1.57.1, so I keep it consistent. To be honest I do not care as long as we increase version.

teclator: Yep, but once unified then it does not make sense to maintain the fourth digit, so just continue increasing the third one, but it could be done when next MU be realeased

jreidinger: agreed, no difference to changelog unless required.

PR #154, opened 2018-03-02
Fixed removing the password expiration date

mvidner:
Merging #153 to master

plus some minimal refactoring of the affected method, which had 1900 lines. And the obviously correct refactoring obviously introduced a stupid bug which I found with manual testing.

Plus Coveralls and Yardoc.

### DIFF
--- a/src/include/users/dialogs.rb
+++ b/src/include/users/dialogs.rb
@@ -155,6 +155,44 @@ module Yast
       ret == :ok ? pw : nil
     end
 
+
+    # helper function: show a popup if existing home directory should be used
+    # and its ownership should be changed
+    # @param dir [String]
+    # @param chown_default [Boolean]
+    def ask_chown_home(dir, chown_default)
+      UI.OpenDialog(
+        Opt(:decorated),
+        HBox(
+          HSpacing(1),
+          VBox(
+            VSpacing(0.2),
+            Label(
+              # popup label, %1 is path to directory
+              Builtins.sformat(_("The home directory (%1) already exists.\nUse it anyway?"), dir)
+            ),
+            Left(
+              # checkbox label
+              CheckBox(Id(:chown_home), _("&Change directory owner"), chown_default)
+            ),
+            ButtonBox(
+              PushButton(Id(:yes), Opt(:default), Label.YesButton),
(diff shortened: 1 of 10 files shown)


src/include/users/dialogs.rb
```diff
@@ -155,6 +155,145 @@ def AskForOldPassword
...
+            ),
+            VSpacing(0.2)
+          ),
+          HSpacing(1)
+        )
+      )
+      ui = UI.UserInput
+      retmap = { "retval" => ui == :yes }
+      if ui == :yes
+        Ops.set(retmap, "chown_home", UI.QueryWidget(Id(:chown_home), :Value))
```
jreidinger: set is not needed here as above it creates solid hash.

src/include/users/dialogs.rb
```diff
@@ -155,6 +155,145 @@ def AskForOldPassword
...
+              PushButton(Id(:yes), Opt(:default), Label.YesButton),
+              PushButton(Id(:no), Label.NoButton)
+            ),
+            VSpacing(0.2)
+          ),
+          HSpacing(1)
+        )
+      )
+      ui = UI.UserInput
+      retmap = { "retval" => ui == :yes }
```
jreidinger: two times identical comparison to `:yes`. Maybe own variable?

src/include/users/dialogs.rb
```diff
@@ -155,6 +155,145 @@ def AskForOldPassword
...
+      min = GetInt(Ops.get(user, "shadowMin"), -1)
+      warn = GetInt(Ops.get(user, "shadowWarning"), -1)
+
+      if last_change != "0"
+        last_change_label = format_days_after_epoch(last_change, "%x")
+      else
+        # label (date of last password change)
+        last_change_label = _("Never")
+      end
+      if expires != "0" && expires != "-1" && expires != ""
```
jreidinger: `if !["0", "-1", ""].include?(expires)`

jreidinger:
in general looks good code wise, but what I miss pretty heavily is any sort of test coverage.

mvidner:
@jreidinger thanks for the comments; updated

PR #155, opened 2018-03-13
fix setting root password via RootPassword boot option (bsc#1081958)

wfeldt:
The old code would encrypt the password twice. Use the plain password in
this place.

### DIFF
--- a/src/modules/Users.pm
+++ b/src/modules/Users.pm
@@ -6304,8 +6304,8 @@ sub Import {
 	    my $pw = Linuxrc->InstallInf ("RootPassword");
 	    if (defined $pw){
 
-		$root_user{"userPassword"}	=
-		    $self->CryptPassword ($pw, "system");
+		y2milestone ("updating root password from install.inf");
+		$root_user{"userPassword"} = $pw;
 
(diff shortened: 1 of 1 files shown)


src/modules/Users.pm
```diff
@@ -6304,8 +6304,8 @@ sub Import {
 	    my $pw = Linuxrc->InstallInf ("RootPassword");
 	    if (defined $pw){
 
-		$root_user{"userPassword"}	=
-		    $self->CryptPassword ($pw, "system");
+		y2milestone ("updating root password from install.inf");
+		$root_user{"userPassword"} = $pw;
```
jreidinger: @wfeldt how do you test it? What is exact workflow? because for me it does not work. I have to add users modified and also with this change password is written plain text to /etc/shadow. I will prepare pr with my fix that for two of my test cases works. Test cases was:

1. no users entry in autoyast xml
1. users entry in autoyast xml but not for root user

jreidinger: Now I found why it works in your case. Customer have in profile root user and specified "encrypted" => false. But if he does not specify it, it is set to true. So in my case it does not work. So I add fix to my pull request, that ensure it always work properly.

PR #157, opened 2018-04-06
Do not crash if root is not defined

imobachgs:
Fixes bsc#1088183.


wfeldt:
BTW, as mls spotted, these two lines

https://github.com/yast/yast-users/blob/master/src/modules/Users.pm#L53-L54

are clearly wrong. `()` should probably be `{}` here. (Compare to similar blobs further below.)

teclator:
LGTM

imobachgs:
Thanks both! We will address what Steffen comments in another PR.

PR #159, opened 2018-04-11
Fix users test

imobachgs:


test/users_test.rb
```diff
@@ -0,0 +1,17 @@
...
+require_relative "test_helper"
+
+Yast.import "Users"
+
+describe "Users" do
+  subject(:users) { Yast::Users }
+
+  describe "#WriteAuthorizedKeys" do
+    context "when no user is defined" do
+      it "does not crash" do
```
lslezak: I think the description should be rather `it "returns true"` because that's what the test actually checks.

The `does not crash` test would rather check that the `NoMethodError` exception is not raised.

imobachgs: Well, returning true is just a trick because it just failed silently in the Perl part and, as nothing was returned, we were not able to check if it was working. But I will add a comment.

PR #160, opened 2018-06-04
Root password

imobachgs:
This PR is a backport of #156.


imobachgs:
@jreidinger Thanks! I've checked that the fix for Linuxrc is already included in the SP4 project in IBS.

PR #166, opened 2018-10-18
read ssh keys from root user only if the user exists (bsc#1112119, bsc#1107456)

wfeldt:
When working with NIS 'root' is typically not included.


teclator:
It seems similar to https://github.com/yast/yast-users/pull/157/files

PR #173, opened 2018-10-31
Allow to set a public key for the root user

imobachgs:
This PR allows to use a public key to authenticate the root user (see [fate#324690](https://fate.suse.com/324690)). The user is allowed to specify a password, a public key or both of them.

## How it looks

select-public-key


package/yast2-users.spec
```diff
@@ -64,6 +66,9 @@ Requires:       yast2-core >= 2.21.0
 
 Requires:       yast2-ruby-bindings >= 1.0.0
 
+# ssh-keygen
+Requires:       openssh
```
lslezak: SSH is a quite basic tool so it will be very likely present even in a small system. But do we need it also in installed system or is this feature present only in installation?

src/lib/users/dialogs/inst_root_first.rb
```diff
@@ -19,69 +19,46 @@
...
     end
 
+  private
+
     # Returns whether we need/ed to create new UI Wizard
-    def separate_wizard_needed?
+    #
+    # @note We do not need to create a wizard dialog in installation, but it's helpful when testing
+    #   all manually on a running system
+    def should_open_dialog?
```
lslezak: The `should_open_dialog?` is already defined in the `CWM::Dialog` (https://github.com/yast/yast-yast2/blob/master/library/cwm/src/lib/cwm/dialog.rb#L43). Why we need a different implementation here?

src/lib/users/leaf_blk_device.rb
```diff
@@ -0,0 +1,121 @@
...
+require "yast2/execute"
+
+module Y2Users
+  # This class represents a block device reported by `lsblk` as a leaf node
+  #
+  # As we cannot use a devicegraph for the time being, this class extracts the
+  # block devices information from the lsblk command. Only the leaf devices
+  # are taken into account and it only includes information which is relevant
+  # for the {Y2Users::Widgets::DiskSelector} widget.
+  class LeafBlkDevice
```
lslezak: Would this class also make sense outside of `yast2-users`? Maybe we could share the class...

src/lib/users/leaf_blk_device.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      # Gets `lsblk` into a Hash
+      #
+      # @return [Hash] Hash containing data from `lsblk`
+      def lsblk
+        output = Yast::Execute.locally(
+          "/usr/bin/lsblk", "--inverse", "--json", "--paths",
+          "--output", "NAME,TRAN,FSTYPE,MODEL", stdout: :capture
+        )
+        JSON.parse(output)
```
lslezak: `Yast::Execute.locally` returns `nil` in case of failure. Then `JSON.parse(nil)` will crash here... :boom:

src/lib/users/ssh_public_key.rb
```diff
@@ -0,0 +1,76 @@
...
+
+require "yast2/execute"
+
+module Y2Users
+  # This class is a simplified representation of a OpenSSH public key.
+  #
+  # @example Read a public key
+  #   key = Y2Users::SSHPublicKey.new(File.read("id_rsa.pub"))
+  #   key.fingerprint # => "SHA256:uadPyDQj9VlFZVjK8UNp57jOnWwzGgKQJpeJEhZyV0I"
+  class SSHPublicKey
```
lslezak: Again, would it make sense to share this class??

src/lib/users/widgets/public_key_selector.rb
```diff
@@ -0,0 +1,282 @@
...
+
+      # UI which shows the public key content
+      #
+      # @return [Yast::Term]
+      def public_key_content
+        VBox(
+          Left(Label(value.fingerprint)),
+          HBox(
+            Left(Label(value.comment)),
+            Right(PushButton(Id(:remove), Opt(:notify), _("Remove")))
```
lslezak: Use `Label::RemoveButton` here (https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Label.rb#L182).

src/lib/users/widgets/public_key_selector.rb
```diff
@@ -0,0 +1,282 @@
...
+      # @return [Yast::Term]
+      def blk_device_selector
+        VBox(
+          Left(
+            HBox(
+              blk_devices_combo_box,
+              PushButton(Id(:refresh), Opt(:notify), _("Refresh..."))
+            )
+          ),
+          Left(PushButton(Id(:browse), Opt(:notify), _("Browse...")))
```
lslezak: Use `Label::BrowseButton` here (https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Label.rb#L236).

src/lib/users/widgets/public_key_selector.rb
```diff
@@ -0,0 +1,282 @@
...
+      # This widget displays includes a list of selectable disk and a button to browse the
+      # selected one.
+      #
+      # @return [Yast::Term]
+      def blk_device_selector
+        VBox(
+          Left(
+            HBox(
+              blk_devices_combo_box,
+              PushButton(Id(:refresh), Opt(:notify), _("Refresh..."))
```
lslezak: Use `Label::RefreshButton` here (https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Label.rb#L188).

src/lib/users/widgets/public_key_selector.rb
```diff
@@ -0,0 +1,282 @@
...
+        )
+      end
+
+      # Key comment to show to the user
+      #
+      # @note When no comment is present, the widget shows 'no comment' just as ssh-keygen does.
+      #
+      # @return [String]
+      def comment_value
+        value.comment || _("no comment")
```
lslezak: This will be hard to translate without the context, please add a `TRANSLATORS:` comment.

src/lib/users/leaf_blk_device.rb
imobachgs: That's a fair point but I would say that we should use libstorage-ng instead. But the problem is that we cannot do the probing again at this point (it will modify the probing/staging devicegraphs). I will open the discussion about introducing some libstorage-ng changes in order to be able to do that, but it will take time.

src/lib/users/ssh_public_key.rb
imobachgs: I would keep it locally and move it when needed (as the `SSHAuthorizedKeys` module).

test/lib/users/widgets/inst_root_first_test.rb
```diff
@@ -0,0 +1,92 @@
...
+      before do
+      end
+
+      it "returns false" do
+        expect(widget.validate).to eq(false)
+      end
+
+      it "displays an error" do
+        expect(Yast2::Popup).to receive(:show)
+          .with(/to provide at least a password/, headline: :error)
```
lslezak: We set the English locale during test, so OK. (https://github.com/yast/yast-users/blob/master/test/test_helper.rb#L25)

package/yast2-users.spec
imobachgs: It is true that we are targeting the installation only... by now. But as part of this feature, we will provide a way to manage public keys to users in an already installed system and we might need this tool.

src/lib/users/dialogs/inst_root_first.rb
imobachgs: Because it is a different condition (I am just keeping the old behaviour to use a new dialog only when `Mode.normal`).

PR #175, opened 2018-11-06
Add Rubocop and fix some documentation issues

imobachgs:


.rubocop.yml
```diff
@@ -0,0 +1,42 @@
+# use the shared YaST defaults
+inherit_from:
+  /usr/share/YaST2/data/devtools/data/rubocop_yast_style.yml
+
+# Redundant returns add legibility for developers used to other languages
```
jreidinger: still looks like copy from storage-ng rubocop. Other yast parts do not need to look like C++

imobachgs: Yes, you are right. I took the storage-ng configuration as the base for this one. Fixed.

.rubocop.yml
```diff
@@ -0,0 +1,39 @@
...
+# resolutions and Github's pull request view. But let's open the threshold a
+# little bit
+Metrics/LineLength:
+  Max: 105
+
+# Enforce if/unless at the end only for really short lines
+Style/IfUnlessModifier:
+  MaxLineLength: 60
+
+# some storage API have size method, but without empty? method
```
jreidinger: does user use storage api with empty?

.rubocop.yml
```diff
@@ -0,0 +1,39 @@
...
+inherit_from:
+  /usr/share/YaST2/data/devtools/data/rubocop_yast_style.yml
+
+# Don't enforce any particular name for block params
+SingleLineBlockParams:
+  Enabled: false
+
+# The general YaST agreement is 100 characters, so it fits into standard
+# resolutions and Github's pull request view. But let's open the threshold a
+# little bit
```
jreidinger: looks like c&p from storage-ng. I do not see much reasoning for having threshold.

.rubocop.yml
```diff
@@ -0,0 +1,39 @@
...
+# Enforce if/unless at the end only for really short lines
+Style/IfUnlessModifier:
+  MaxLineLength: 60
+
+# some storage API have size method, but without empty? method
+# for details see https://github.com/yast/yast-storage-ng/pull/83
+Style/ZeroLengthPredicate:
+  Enabled: false
+
+Style/FormatString:
```
jreidinger: rationale?

.rubocop.yml
```diff
@@ -0,0 +1,39 @@
...
+  Exclude:
+    - 'src/lib/users/encryption_proposal.rb'
+    - 'src/lib/users/proposal.rb'
+
+Style/FormatString:
```
jreidinger: duplicite

src/lib/users/ssh_authorized_keyring.rb
```diff
@@ -62,26 +62,34 @@ def default_message
 
       # The home directory does not exist.
       class HomeDoesNotExist < PathError
-        # @return default_message [String] Default error message
-        def default_message; "Home directory does not exist" end
+        # @return [String] Default error message
+        def default_message
+          "Home directory does not exist"
```
jreidinger: no translation? Do we show it to users or not?

imobachgs: No, we just log this information. See https://github.com/yast/yast-users/blob/a90550d5b483982b93aa88fc7dafa9f24fd96f3d/src/modules/SSHAuthorizedKeys.rb#L51.

src/lib/users/encryption_method.rb
```diff
@@ -55,12 +55,22 @@ class NotFoundError < Error
...
       "sha512" => N_("SHA-512")
-    }
+    }.freeze
     private_constant :LABELS
 
-    DEFAULT_ID = "sha512"
+    DEFAULT_ID = "sha512".freeze
     private_constant :DEFAULT_ID
 
+    class << self
```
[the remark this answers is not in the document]
imobachgs: AFAIK, this method was not private at all.

```
irb(main):001:0> class MyClass
irb(main):002:1>   private
irb(main):003:1>   def self.my_method
irb(main):004:2>     puts "hi!"
irb(main):005:2>   end
irb(main):006:1> end
=> :my_method
irb(main):007:0> MyClass.my_method
hi!
```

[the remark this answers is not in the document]
jreidinger: yes, it is true, but what I write works. So I know it does not work before. I just propose alternative.

PR #176, opened 2018-11-07
Run also "rake pot" at Travis

lslezak:
- Run also `rake pot` at Travis to check for possible translation issues (e.g. missing textdomain)
- Here we use the `yast-travis-cpp` script which does not run the pot check


lslezak:
Actually we should use a generic solution, see https://github.com/yast/docker-yast-cpp/pull/21

PR #192, opened 2018-12-20
Hardening

jreidinger:


src/include/users/cmdline.rb
```diff
@@ -390,7 +390,7 @@ def UserDeleteHandler(options)
       out = Convert.to_map(
         SCR.Execute(
           path(".target.bash_output"),
-          Builtins.sformat("ps --no-headers -u %1", uid)
+          Builtins.sformat("/usr/bin/ps --no-headers -u %1", uid.to_i)
```
mvidner: Why to_i?

imobachgs: If uid cannot be converted to a proper number, the result will be `0` which, after all, it is the root's UID. In this case, as YaST runs as root, the only potential problem is that detecting whether the user is logged in or not could fail (and it will prevent deleting the user).

For me it is fine, but I just wanted to point it out.

jreidinger: yes, it is true, but it should not be security issue and also code above handles it ( so it "should" not happen ). This is mainly last defense against malformed input.

src/clients/users_plugin_ldap_shadowaccount.rb
```diff
@@ -121,17 +123,9 @@ def main
...
-                Builtins.sformat(
-                  "date --date='1970-01-01 00:00:01 %1 days' ",
-                  @expires
-                ),
-                "+\"%Y-%m-%d\""
-              )
-            )
+          @out = SCR.Execute(
+            path(".target.bash_output"),
+            "/usr/bin/date --date='1970-01-01 00:00:01 %1 days' #{@expires.to_s.shellescape}+\"%Y-%m-%d\""
```
imobachgs: This interpolation looks wrong. Should not be `%1` replaced with `@expires.to_s.shellescape`=

src/clients/users_plugin_ldap_shadowaccount.rb
```diff
@@ -258,14 +252,9 @@ def main
...
-                  SCR.Execute(
-                    path(".target.bash_output"),
-                    Ops.add(
-                      Builtins.sformat("date --date='%1 UTC' ", @exp_date),
-                      "+%s"
-                    )
-                  )
+                @out = SCR.Execute(
+                  path(".target.bash_output"),
+                  "/usr/bin/date --date='%1 UTC' #{@exp_date.to_s.shellescape}+%s"
```
imobachgs: Same than above.

src/include/users/dialogs.rb
```diff
@@ -1576,14 +1578,9 @@ def EditUserDialog(what)
...
-                SCR.Execute(
-                  path(".target.bash_output"),
-                  Ops.add(
-                    Builtins.sformat("date --date='%1 UTC' ", exp_date),
-                    "+%s"
-                  )
-                )
+              out = SCR.Execute(
+                path(".target.bash_output"),
+                "/usr/bin/date --date=#{exp_date.shellescape}%1' UTC' +%s"
```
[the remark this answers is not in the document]
jreidinger: good catch, I forgot to remove %1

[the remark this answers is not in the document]
jreidinger: just `''` is placed wrong in you expectation as shellescape have to not be enclosed in quotes. So it should be something like `"/usr/bin/date --date=#{exp_date.shellescape}' UTC' +%s"`

src/include/users/widgets.rb
```diff
@@ -1830,17 +1832,9 @@ def InitDefaults(key)
...
-              Builtins.sformat(
-                "date --date='1970-01-01 00:00:01 %1 days' ",
-                expire
-              ),
-              "+\"%Y-%m-%d\""
-            )
-          )
+        out = SCR.Execute(
+          path(".target.bash_output"),
+          "/usr/bin/date --date='1970-01-01 00:00:01 '#{expire.shellescape}' days' +\"%Y-%m-%d\""
```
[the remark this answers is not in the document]
jreidinger: It have to be there, as shellescape have to *not* be enclosed in quotes, otherwise it produce wrong result, so I intentionally end quote and start it again after interpolation.

src/include/users/widgets.rb
```diff
@@ -1881,14 +1875,9 @@ def StoreDefaults(key, event)
...
-          SCR.Execute(
-            path(".target.bash_output"),
-            Ops.add(
-              Builtins.sformat("date --date='%1 UTC' ", new_exp_date),
-              "+%s"
-            )
-          )
+        out = SCR.Execute(
+          path(".target.bash_output"),
+          "/usr/bin/date --date=#{new_exp_date.shellescape}' UTC' +%s"
```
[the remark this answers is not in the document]
jreidinger: no, it is intention. See above.

PR #193, opened 2019-01-08
Fix syntax

mvidner:
Fixup of #192 "Hardening"

- [x] changes and version
- [x] catch this automatically via`perl -c`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@ FROM yastdevel/cpp
 RUN zypper --gpg-auto-import-keys --non-interactive in --no-recommends \
   cracklib-devel \
   perl-Digest-SHA1 \
+  perl-X500-DN \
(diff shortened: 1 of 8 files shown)


jreidinger:
looks also good for me.

mvidner:
@lslezak I suggest that we merge this only after you add the `perl -wc`, to test that it really catches the bug.

lslezak:
JFYI: The Perl syntax check has been added in https://github.com/yast/docker-yast-cpp/pull/23 (and https://github.com/yast/docker-yast-ruby/pull/32).

lslezak:
The Perl syntax check now works at Travis, see the output log: https://travis-ci.org/yast/yast-users/builds/477217161#L1017. (I just had to add the `perl-X500-DN` package to the install list.)

PR #194, opened 2019-01-11
Fix bsc 1121473

imobachgs:
Fixes [bsc#1121473](https://bugzilla.suse.com/show_bug.cgi?id=1121473).

Additionally, it adds a unit test to avoid regressions and fixes the test name.


imobachgs:
Thanks!

PR #198, opened 2019-02-27
Delete remaining dead code for encrypted homes

mvidner:
- https://trello.com/c/plarcsbX/770-1-sle15-sp1-no-mkdir-is-not-in-usr-sbin
- https://bugzilla.suse.com/show_bug.cgi?id=1127138
- https://fate.suse.com/323541


mvidner:
@schubi2 do you want to review this?

PR #202, opened 2019-04-12
- don't copy skel files for system users (bsc#1130158)

wfeldt:
This patch fixes three issues:

1. autoyast imported system users still got a copy of /etc/skel, which is nonsense and can break things
2. the decision on whether to touch a home dir at all is based on whether the home dir is empty or not - but the implementation of the detection of this situation was wrong
3. added more logging as the current log verbosity is really insufficient

Some more remarks:

The observed behavior differed between xfs (or ext4) and btrfs due to point 2. above. E.g. stat(DIR) returns 0 for empty btrfs directories while it is !=0 for empty xfs dirs.

Note that there is even a user `rtkit` that has `/proc` as home dir. As `stat("/proc")` gives 0, the old code tried to modify all `/proc` entries.

### DIFF
--- a/src/modules/Users.pm
+++ b/src/modules/Users.pm
@@ -4475,9 +4475,17 @@ sub Write {
 		if ($user_mod eq "imported" || $user_mod eq "added") {
 		    y2usernote ("User '$username' created");
 
+                    # *** FIXME ***
+                    #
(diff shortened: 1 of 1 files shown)


schubi2:
This has gone to:
https://github.com/yast/yast-users/pull/203

PR #203, opened 2019-04-15
Improved checking of already existing home

schubi2:
This is Steffen's PR:
https://github.com/yast/yast-users/pull/202/files
I have added the package stuff and have tested it again.

### DIFF
--- a/src/modules/Users.pm
+++ b/src/modules/Users.pm
@@ -4493,8 +4493,6 @@ sub Write {
 		$chown_home	        = 1 if (!defined $chown_home);
 		my $skel	        = $useradd_defaults{"skel"};
 
-		y2milestone ("User = ", Dumper(\%user));
-
 		if ($user_mod eq "imported" || $user_mod eq "added") {
 		    y2usernote ("User '$username' created");
 
@@ -4566,10 +4564,13 @@ sub Write {
 		    }
 		}
 
-			# Write authorized keys to user's home (FATE#319471)
-      if (defined($user{"authorized_keys"})) {
-				SSHAuthorizedKeys->write_keys($home, $user{"authorized_keys"});
-      }
+		# Write authorized keys to user's home (FATE#319471)
+		if (defined($user{"authorized_keys"})) {
+		    SSHAuthorizedKeys->write_keys($home, $user{"authorized_keys"});
+		}
+
+		$user{"userPassword"} = "*****"; # Reset before logging
+		y2milestone ("User = ", Dumper(\%user));
 	    }
 	}
     }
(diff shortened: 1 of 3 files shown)


src/modules/Users.pm
```diff
@@ -4466,12 +4493,14 @@ sub Write {
 		$chown_home	        = 1 if (!defined $chown_home);
 		my $skel	        = $useradd_defaults{"skel"};
 
+		y2milestone ("User = ", Dumper(\%user));
```
lslezak: BTW what does this `%user` data actually contain? We need to be careful, if that contains the password we should not write it into the log.

src/modules/Users.pm
```diff
@@ -250,6 +254,29 @@ YaST::YCP::Import ("SSHAuthorizedKeys");
 ##-------------------------------------------------------------------------
 ##----------------- various routines --------------------------------------
 
+# Check whether the argument is a directory and is not empty.
+#
+# Return 1 (non-empty dir) or 0.
+#
+# Note: do NOT use FileUtils->GetSize() for this!
```
lslezak: I'm just curions, why not? What's the reason? Maybe it should be written here...

schubi2: Steffen has explained it:
..
2. the decision on whether to touch a home dir at all is based on whether the home dir is empty or not - but the implementation of the detection of this situation was wrong
..

The observed behavior differed between xfs (or ext4) and btrfs due to point 2. above. E.g. stat(DIR) returns 0 for empty btrfs directories while it is !=0 for empty xfs dirs.

src/modules/Users.pm
```diff
@@ -4566,10 +4564,13 @@ sub Write {
...
-      if (defined($user{"authorized_keys"})) {
-				SSHAuthorizedKeys->write_keys($home, $user{"authorized_keys"});
-      }
+		# Write authorized keys to user's home (FATE#319471)
+		if (defined($user{"authorized_keys"})) {
+		    SSHAuthorizedKeys->write_keys($home, $user{"authorized_keys"});
+		}
+
+		$user{"userPassword"} = "*****"; # Reset before logging
+		y2milestone ("User = ", Dumper(\%user));
```
lslezak: Should we restore back the original password? I mean the code executed later will see the password placeholder instead od the real password. That might be confusing or can cause bugs. I'd rather restore it back, just to be on the safe side and avoid surprises later...

src/modules/Users.pm
lslezak: Oh, I see, so it is FS dependent... :open_mouth:

PR #205, opened 2019-04-16
Offer autologin only if a display manager that supports it is installed

shundhammer:
# Trello

https://trello.com/c/VVM2ib4G/

# Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1128385 (openSUSE)
https://bugzilla.suse.com/show_bug.cgi?id=1060173 (SLES)

# Related PR

https://github.com/yast/yast-pam/pull/13

# Problem

During installation and later in the installed system, "autologin" is offered even if it doesn't make sense and if it can't even work: If a user selects the "server" role or some other package combination that does not provide X11 or Wayland or, in general, a graphical login, the "Autologin" checkbox is available.

Selecting that option has no effect in that case.

# Fix

Don't only rely on the control.xml file that has a parameter if that should be offered at all; also check if the capability is available, i.e. if a display manager (login manager) that supports that feature is or will be installed.

`Autologin.supported?` from the _yast2-pam_ package (see related PR https://github.com/yast/yast-pam/pull/13) provides that information.

# Test

## Installed System

- Start "yast2 users" and

  - Check /var/log/YaST2/y2log for "Autologin is supported" or "Autologin is not supported"
  - On the main dialog of the users module, check the "Expert Options" menu whether or not there is a menu entry "Login Settings" (this is only about autologin)

You might need to modify the package list in `/usr/share/YaST2/modules/Autologin.rb` to test the other case.

## inst-sys

- Start the installation
- Select system role "Server" (or another role without graphical desktop)
- Continue to the "Create First User" dialog
- Check if the "Autologin" checkbox in that dialog is really disabled

- Go back to the role selection
- Select another role, this time _with_ a graphical desktop
- Continue to the "Create First User" dialog
- Check if the "Autologin" checkbox is enabled now

### System Role "Server (text mode)" Selected

inst-first-user-server-role

### System Role "GNOME Desktop" Selected

inst-first-user-gnome-role

### DIFF
--- a/package/yast2-users.spec
+++ b/package/yast2-users.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-users
-Version:        4.1.12
+Version:        4.1.13
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
--- a/package/yast2-users.spec
+++ b/package/yast2-users.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-users
-Version:        4.1.13
+Version:        4.2.0
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
@@ -46,7 +46,7 @@ Requires:       perl-gettext
 Requires:       yast2-country
 
 # Autologin.supported?
-Requires:       yast2-pam >= 4.1.1
+Requires:       yast2-pam >= 4.2.0
 
 Requires:       yast2-security
 Obsoletes:      yast2-users-devel-doc
--- a/package/yast2-users.spec
+++ b/package/yast2-users.spec
@@ -44,7 +44,10 @@ Requires:       perl-Digest-SHA1
 Requires:       perl-X500-DN
 Requires:       perl-gettext
 Requires:       yast2-country
-Requires:       yast2-pam
+
+# Autologin.supported?
+Requires:       yast2-pam >= 4.1.1
+
 Requires:       yast2-security
 Obsoletes:      yast2-users-devel-doc
 Conflicts:      autoyast2 < 3.1.92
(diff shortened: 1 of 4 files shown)


package/yast2-users.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-users
-Version:        4.1.12
+Version:        4.1.13
```
lslezak: Similar to https://github.com/yast/yast-pam/pull/13#pullrequestreview-227263255, very likely `4.2.0` should be here.

lslezak:
Thinking about the UX aspect, maybe we could use a label like `Automatic Login (not supported by the current package selection)` or something like that when the check box is disabled. It's quite confusing when a widget is disabled and there is no apparent reason why, esp. if it depends on the package selection which is configured elsewhere.

shundhammer:
Answered in IRC:

```
[17:25:20] <HuHa> lslezak: well, that would be most unusual - we never ever do that anywhere else
[17:25:42] <HuHa> lslezak: once a long time ago (in my Heitec time) I had that idea of providing that kind of information in general
[17:26:02] <HuHa> i.e. when you disable a widget programmatically, you'd have to add a reason why
[17:26:08] <HuHa> and that reason would be in a tooltip
[17:26:21] <HuHa> of course that would mean that you might have to add multiple reasons
[17:26:47] <HuHa> "disabled because:  - you are too stupid   - this doesn't work    - you don't deserve that"   ;-)
[17:27:18] <HuHa> but adding that text to the widget label is not a good idea IMHO
[17:27:36] <HuHa> for starters, because it would grow out of bounds for NCurses pretty quickly
[17:27:57] <HuHa> and when we start doing that, people will quickly demand we do it all over the place, not just there
[17:28:11] <lslezak> HuHa: ok
[17:28:13] <HuHa> where in this case it should be obvious enough to the user why he can't do it
[17:28:22] <HuHa> because for a server, that wouldn't work
[17:28:30] <HuHa> that's where the bug report comes from BTW ;-)
[17:29:41] <lslezak> HuHa: I know, nobody reads the help, but what about adding a note there like "the checkbox is disabled when the selected/installed display manager does not support autologin" ?
[17:29:56] <HuHa> IMHO that would just confuse the users more
[17:30:12] <HuHa> in the normal case it will be quite obvious why it's disabled
```

PR #209, opened 2019-06-19
Fixed unit test: skip Package::Installed() call (bsc#1138668)

lslezak:
## Problem

- Failing old test suite
```
[   92s]  Write	.target.string "/etc/shadow" "at:!:13636:0:99999:7:::\nbin:*:13636::::::\ndaemon:*:13636::::::\nmail:*:13636::::::\nnobody:*:13636::::::\nroot:password:13636::::::\nuucp:*:13636::::::\nhh:heslo:13727:0:99999:7:::\n+::0:0:0::::\n" true
[   92s] -Execute	.target.bash "/usr/sbin/nscd -i passwd" 0
[   92s]  Return
```

- It turned out that it depends whether the `nscd` package is installed during the RPM build or not

## Fix

- Check `Mode.test` and do not check for the package presence in that case
- 4.2.3

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jun 19 07:39:49 UTC 2019 - Ladislav Slezák <lslezak@suse.cz>
+
+- Fixed failing old testsuite: do not depend on the environment,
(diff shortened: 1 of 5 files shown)


teclator:
Just for record: The issue is consequence of https://github.com/yast/yast-yast2/pull/938

PR #219, opened 2019-11-21
Disable block device selector in WSL (boo#1154238)

lnussel:


lnussel:
needs https://github.com/yast/yast-yast2/pull/985

teclator:
As commented in https://github.com/yast/yast-yast2/pull/985#issuecomment-557518977, would be nice to have some description, links etc in order to review it.

lnussel:
https://github.com/yast/yast-yast2/pull/990 is in so this one should be ready too

src/lib/users/widgets/public_key_selector.rb
```diff
@@ -234,20 +242,25 @@ def update
...
-          mounted = Yast::SCR.Execute(
-            Yast::Path.new(".target.mount"), [selected_blk_device_name, dir], "-o ro"
-          )
-          if mounted
-            read_key_from(dir)
-          else
-            report_mount_error(selected_blk_device_name)
+        if Yast::Arch.is_wsl
+          # /mnt has windows drives mounted by default
+          read_key_from("/mnt")
```
lslezak: It's better to use `Yast::Installation.destdir` instead of the hardcoded `/mnt`. (You might need to add `Yast.import "Installation"`.)

src/lib/users/widgets/public_key_selector.rb
```diff
@@ -213,8 +219,10 @@ def blk_devices_combo_box
       #
       # @return [Array<LeafBlkDevice>] List of devices
       def available_blk_devices
-        @available_blk_devices ||= LeafBlkDevice.all.select do |dev|
-          dev.filesystem? && dev.fstype != :squash && dev.transport?
+        if !Yast::Arch.is_wsl
```
lslezak: On WSL this returns `nil`, is that OK? I think this would be safer:

```ruby
return [] if Yast::Arch.is_wsl
```

package/yast2-users.changes
```diff
@@ -6,6 +12,12 @@ Mon Nov 25 15:37:16 UTC 2019 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
   - Write login.defs configuration to /etc/login.defs.d/.
 - 4.2.6
 
+-------------------------------------------------------------------
+Thu Nov 21 16:15:00 UTC 2019 - Ludwig Nussel <lnussel@suse.de>
+
+- Disable block device selector in WSL (boo#1154238)
+- 4.2.6
```
jreidinger: this looks wrong. It should not be there

src/lib/users/widgets/public_key_selector.rb
jreidinger: better to not return nil. So something like
```
return [] if Yast::Arch.is_wsl

... old code
```

src/lib/users/widgets/public_key_selector.rb
```diff
@@ -234,20 +242,25 @@ def update
...
-        begin
-          mounted = Yast::SCR.Execute(
-            Yast::Path.new(".target.mount"), [selected_blk_device_name, dir], "-o ro"
-          )
-          if mounted
-            read_key_from(dir)
-          else
-            report_mount_error(selected_blk_device_name)
+        if Yast::Arch.is_wsl
+          # /mnt has windows drives mounted by default
```
jreidinger: any link for more details how wsl works e.g. what is where?

src/lib/users/widgets/public_key_selector.rb
jreidinger: and same stylish thing. You can early quit with 
```
# /mnt has windows drives mounted by default
return read_key_from("/mnt") if Yast::Arch.is_wsl

... old code
```

jreidinger:
and btw some test coverage would be nice, so we do not break it in near future.

package/yast2-users.changes
lslezak: This entry looks wrong, probably a wrongly resolved conflict...

src/lib/users/widgets/public_key_selector.rb
lnussel: it's unrelated to destdir of an installation actually. MS decided to use /mnt for mounting Windows drives. See eg https://docs.microsoft.com/de-de/windows/wsl/faq

src/lib/users/widgets/public_key_selector.rb
lnussel: it's all on microsoft.com :/ I'd refer to https://docs.microsoft.com/de-de/windows/wsl/faq for example. Not much special there, just /mnt used by Windows to mount Windows drives by default.

package/yast2-users.changes
lnussel: indeed

src/lib/users/widgets/public_key_selector.rb
lnussel: did that

src/lib/users/widgets/public_key_selector.rb
lnussel: i've changed the code to not need that at all, thanks!

package/yast2-users.changes
lnussel: indeed

lnussel:
> and btw some test coverage would be nice, so we do not break it in near future.

Sure but beyond my abilities

PR #224, opened 2020-02-05
Improve readme

Firstyear:
This improves the readme of the git repo to help new developers get
started with the project quicker.


ancorgs:
I just came across this pull request while checking old stuff.

Most things explained here are not specific to yast2-users, but are notes about YaST development in general. From that point of view, it would make more sense to link to https://yast.opensuse.org/contributing

There is no chance we can keep the same development information up to date in all the 100+ READMEs of the individual YaST repositories.

PR #228, opened 2020-05-13
Fixed crash in AY user proposal

schubi2:
Problem
========
* Showing AY user settings in the proposal screen crashed when the user would like to change it.
* The regarding UI does not fit to the settings of the AY configuration file.

Solution
=======
* Fixed crash
* Only showing the settings. The user cannot change it.
users

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed May 13 14:04:35 CEST 2020 - schubi@suse.de
+
+- AY proposal: Fixed crash. Showing user settings only because the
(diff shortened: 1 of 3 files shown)


package/yast2-users.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed May 13 14:04:35 CEST 2020 - schubi@suse.de
+
+- AY proposal: Fixed crash. Showing user settings only because the
+  UI does not fit to the AY settings (bsc#1170815)
```
jreidinger, line 5: I do not get much this changelog from user POV. What is changed for users? Showing in autoyast user proposal to propose only users and groups?

PR #230, opened 2020-05-27
Drop the old/legacy testsuite

dgdavid:
The legacy testsuite, which makes testing by output to log, is a blocker for the evolution of YaST and is causing breakage in Travis. For that reason, we have been trying to migrate it as much as possible to RSpec in the rest of YaST modules. However, this module is mainly written in Perl and we didn't find a good Perl-mocking tool to make that testsuite less fragile and more useful.

Additionally, looking at the history of changes it can be seen that those tests (and the related code) has been mostly untouched for ages, which helps to make the decision of getting rid of them and move on.

So, this PR deletes the old testsuite. Of course, it can be checked if needed at any time thanks to the version control history and ideally every new change from now on should be tested with RSpec, although for that we still needing either, to migrate the existing Perl code to Ruby or find out the way to mock it properly in order to use RSpec for writing the tests.


package/yast2-users.spec
```diff
@@ -39,7 +39,6 @@ BuildRequires:  yast2-core-devel
 BuildRequires:  yast2-devtools >= 4.2.2
 BuildRequires:  yast2-perl-bindings
 BuildRequires:  yast2-security
-BuildRequires:  yast2-testsuite
```
ancorgs, line 42: I suspect perl-XML-Writer can also be removed now from the BuildRequires

dgdavid: I agree. In fact, I was checking the dependencies right now.

Thanks for the _confirmation_ :)

ancorgs:
Among other things, we are doing this in order to bring compatibility with Ruby 2.7, since this is currently failing to build in Staging:J
https://build.opensuse.org/project/show/openSUSE:Factory:Staging:J

That likely deserves a changelog entry and a version bump. Although I'm not sure whether we have a bug number for that.

PR #232, opened 2020-06-25
Do not export system users/groups if target is set to 'compact'

imobachgs:
Honor the `target` argument by not exporting system users/groups if it is set to `compact`. See [AutoYaST2 PR](https://github.com/yast/yast-autoinstallation/pull/631) for further information.

Changes:

* Add an optional "target" argument to the Users.Export method. I am not sure whether the types annotation is right.
* Rewrite the auto client using `Installation::AutoClient` as the base class.


src/lib/users/clients/auto.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      # @note This code is not executed during autoinstallation (instead, the
+      # users_finish is used). However, it is used when running ayast_setup.
+      def write
+        Yast::Users.SetWriteOnly(true)
+        progress_orig = Yast::Progress.set(false)
+        ret = Yast::Users.Write == ""
+        Yast::Progress.set(progress_orig)
+        # Return if configuration  was changed
+        # return boolean
```
jreidinger, line 80: can you just add it to yardoc?

PR #233, opened 2020-07-14
Use new krb5 sbin path 

joseivanlopez:
## Problem

The package *krb5* was adapted to provide binaries at */usr/sbin* path instead of */usr/lib/mit/sbin*, see https://build.opensuse.org/request/show/814662.

YaST is still using the old */usr/lib/mit/sbin* paths. 

* https://trello.com/c/PxH6LJhf/1951-3-new-krb5-path-in-tw

* https://bugzilla.opensuse.org/show_bug.cgi?id=1174078

## Solution

Adapt YaST code to use the new paths.

Note: the changes in the *kbr5* package were done without a change in the version number, so *yast2-users* cannot indicate a conflict with any *kbr5* old version. And also note that this new *yast2-users* package will be submitted to both: SLE-15-SP3 and Factory, but most likely SLE-15-SP3 will not receive any update for *kbr5*. Because that, *yast2-users* has to deal with both possible paths (/usr/lib/mit/sbin and /usr/sbin).

### DIFF
--- a/src/modules/UsersPluginKerberos.pm
+++ b/src/modules/UsersPluginKerberos.pm
@@ -79,6 +79,18 @@ sub contains {
     return 0;
 }
 
+# internal function:
+# check the path of kadmin.local binary
+#
+# Note that the lastest kbr5 package provides the kadmin.local binary at /usr/sbin,
+# but older kbr5 uses /usr/lib/mit/sbin path.
+sub kadmin_path {
+    my $path = "/usr/sbin/kadmin.local";
+    my $old_path = "/usr/lib/mit/sbin/kadmin.local";
+
+    return SCR->Execute (".target.stat", $path) ? $path : $old_path;
+}
+
 ##------------------------------------------
 ##--------------------- global API functions
 
@@ -150,7 +162,7 @@ sub PluginPresent {
 	y2debug ("Kerberos plugin not present");
 	return 0;
     }
-    my $out	= SCR->Execute (".target.bash_output", "/usr/lib/mit/sbin/kadmin.local -nq 'list_principals ".String->Quote("".$data->{uid})."*' | /usr/bin/grep '".String->Quote("".$data->{uid})."*'");
+    my $out	= SCR->Execute (".target.bash_output", "${\kadmin_path()} -nq 'list_principals ".String->Quote("".$data->{uid})."*' | /usr/bin/grep '".String->Quote("".$data->{uid})."*'");
     if ($out->{"stdout"} =~ /^$data->{uid}/ ) {
 	y2milestone ("Kerberos plugin present");
 	return 1;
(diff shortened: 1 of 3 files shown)


src/modules/UsersPluginKerberos.pm
```diff
@@ -150,7 +150,16 @@ sub PluginPresent {
...
 	return 0;
     }
-    my $out	= SCR->Execute (".target.bash_output", "/usr/lib/mit/sbin/kadmin.local -nq 'list_principals ".String->Quote("".$data->{uid})."*' | /usr/bin/grep '".String->Quote("".$data->{uid})."*'");
+
+    # Note that the last kbr5 package provides the kadmin.local binary at /usr/sbin,
+    # but older kbr5 uses /usr/lib/mit/sbin path.
+
+    my $out = SCR->Execute (".target.bash_output", "/usr/sbin/kadmin.local -nq 'list_principals ".String->Quote("".$data->{uid})."*' | /usr/bin/grep '".String->Quote("".$data->{uid})."*'");
+
+    if ($out->{"stdout"} =~ /^$data->{uid}/ ) {
```
jreidinger, line 159: this looks kind of strange to check stdout...why not exit code or stderr? Exit code is 127 if command not found.

src/modules/UsersPluginKerberos.pm
```diff
@@ -258,11 +268,23 @@ sub Write {
...
+      if ( $failure ) {
+	$failure = 0;
+	$pid = open3(\*IN, \*OUT, \*ERR, "$old_command") or do {
+          $failure = 1;
+	};
+      }
+
+      if ( $failure ) {
+	$error = __("Cannot execute kadmin.local.");
+        return YaST::YCP::Boolean (0);
```
jreidinger, line 285: in general this solution looks quite hacky for me. Why not during init just check by e.g. file stat if file exists?

wfeldt: I agree with Josef. Why not simply using `-x $command` to determine which command to use, then run it.

src/modules/UsersPluginKerberos.pm
wfeldt: The same as below. Why not check for the tool first, then run it.

src/modules/UsersPluginKerberos.pm
```diff
@@ -150,7 +150,16 @@ sub PluginPresent {
 	y2debug ("Kerberos plugin not present");
 	return 0;
     }
-    my $out	= SCR->Execute (".target.bash_output", "/usr/lib/mit/sbin/kadmin.local -nq 'list_principals ".String->Quote("".$data->{uid})."*' | /usr/bin/grep '".String->Quote("".$data->{uid})."*'");
+
+    # Note that the last kbr5 package provides the kadmin.local binary at /usr/sbin,
```
wfeldt, line 154: last -> latest

wfeldt:
I would have avoided this SCR beast at all and gone for perl's stat via '-x', though.

PR #234, opened 2020-07-24
Use yast2-pam/Nsswitch module to read NSS configuration

dgdavid:
## Problem

The old SCR nsswitch agent has been replaced by a CFA class able to deal with the new layout having the configuration split between `/usr/etc` and `/etc.` See https://github.com/yast/yast-pam/pull/20 for more details.

## Solution

Start using the [yast2-pam/Nsswitch module](https://github.com/yast/yast-pam/blob/43201e3ed0d50f8dce3a853aef3057265d8aa471/src/modules/Nsswitch.rb) instead of the no longer available SCR agent.

## Test

* **Not unit testing** for the complex Perl module.
* Tested manually running.

### DIFF
--- a/src/modules/Users.pm
+++ b/src/modules/Users.pm
@@ -5753,7 +5753,7 @@ sub ReadNISAvailable {
 
     my $passwd_source = Nsswitch->ReadDb ("passwd");
 
-    if ( grep ( /^nis|compat$/, @$passwd_source ) ) {
+    if ( grep ( /^(nis|compat)$/, @$passwd_source ) ) {
       return (Package->Installed ("ypbind") && Service->Status ("ypbind") == 0);
     }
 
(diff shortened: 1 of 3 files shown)


src/modules/Users.pm
```diff
@@ -5750,15 +5751,12 @@ sub ReadNISMaster {
...
-    my $passwd_source = SCR->Read (".etc.nsswitch_conf.passwd");
-    if (defined $passwd_source) {
-	foreach my $source (split (/ /, $passwd_source)) {
+    my $passwd_source = Nsswitch->ReadDb ("passwd");
 
-	    if ($source eq "nis" || $source eq "compat") {
-		return (Package->Installed ("ypbind") && Service->Status ("ypbind") == 0);
-	    }
-	}
+    if ( grep ( /^nis|compat$/, @$passwd_source ) ) {
```
ancorgs, line 5756: I think this should be `/^(nis|compat)$/`. As it is now, I think these strings satisfies the regexp: "whatevercompat", "niswhatever". But it's regular expressions, so I could be of course wrong.

dgdavid: No, you are not wrong. You're perfectly right. Thanks for catching it :)

PR #237, opened 2020-11-11
Importing public SSH key in administrator screen: Disable browser but…

schubi2:
…ton if no block device is available (bsc#1178509).


wfeldt (approved):
lgtm

jreidinger (approved):

PR #238, opened 2020-12-02
Public key selector: filter is based on filesystems

imobachgs:
It is allowed to use any device that contains a filesystem if such a filesystem is not squashfs (which is basically used by the installer). It should fix [bsc#1178509](https://bugzilla.suse.com/show_bug.cgi?id=1178509).


schubi2 (approved):
LGTM

schubi2:
Final LGTM :-) Thank you !!!

PR #239, opened 2020-12-03
GitHub actions

jreidinger:
- Move from Travis to GitHub Actions
- See examples/templates in https://github.com/yast/.github/pull/3, https://github.com/yast/yast-installation/pull/896


.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+    container: registry.opensuse.org/yast/head/containers/yast-ruby:latest
+
+    steps:
+
+    - name: Git Checkout
+      uses: actions/checkout@v2
+
+    - name: POT Check
+      run: rake check:pot
+
```
lslezak, line 96: As some parts are written in Perl we should add a [Perl syntax check](https://github.com/yast/.github/blob/master/workflow-templates/ci-ruby.yml#L86-L87) here.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+          yast2-perl-bindings
+
+    - name: Package Build
+      run: yast-ci-cpp
+      env:
+        # the tests require specific locale settings to pass
+        LANG: POSIX
+        LC_ALL: ""
+        # for the Perl syntax check
+        PERL5LIB: "./agents-perl/lib"
```
lslezak, line 71: This should be probably moved to the Perl syntax check step, see below.

.github/workflows/ci.yml
jreidinger: see below

.github/workflows/ci.yml
jreidinger: it is done as part of yast-ci-cpp. If we do it below, we have it also below force installation of that perl packages, otherwise perl check failed. So it I think it is better to have it in this one task which also include packages needed for build.

.github/workflows/ci.yml
jreidinger: egh, above :)

.github/workflows/ci.yml
lslezak: Ah, I see, then it makes sense to run it here...

lslezak (approved):

PR #240, opened 2021-03-31
Copy skel from usr

joseivanlopez:
### Problem

*yast-users* does not use the *useradd* command for creating a new user. Instead of that, it does all the work by its own (create home directory, adapt permissions, set password, etc). One of the steps performed by *yast-users* is to execute the *useradd.local* script. That script was taking care of copying the content of */usr/etc/skel* into the new created home directory. But *useradd.local* does not do it anymore. Such a logic was moved to *useradd* in the *shadow* suite.  So now, when a new user is created with *yast-users*, the content of */usr/etc/skel* is not copied at all, resulting on incomplete home directories. 

That missing files in the new home directories is blocking the lastest changes for the *shadow* suite in Factory (including the management of */usr/etc/skel* directly by *useradd*): https://build.opensuse.org/request/show/872327.

* https://bugzilla.suse.com/show_bug.cgi?id=1183136

### Solution

Ideally, *yast-users* should be adapted to use *useradd* command. But that would require to almost re-write the whole module. For now, this issue was fixed by simply copying */usr/etc/skel* before copying */etc/skel* in order to unblock the submit request for the *shadow* suite.

Note that the YaST team has already started working in a new *yast-users-ng* module, which will make use of *useradd* command.

### Tests

* Manually tested.

### DIFF
--- a/src/modules/UsersRoutines.pm
+++ b/src/modules/UsersRoutines.pm
@@ -50,6 +50,9 @@ YaST::YCP::Import ("String");
 # path to btrfs
 my $btrfs = "/usr/sbin/btrfs";
 
+# path to the usr skel
+my $usr_skel = "/usr/etc/skel";
+
 ##-------------------------------------------------------------------------
 ##----------------- helper routines ---------------------------------------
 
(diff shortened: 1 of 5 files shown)


src/modules/UsersRoutines.pm
```diff
@@ -60,23 +63,45 @@ sub btrfs_subvolume {
     return ( SCR->Execute( ".target.bash", $cmd ) eq 0 );
 }
 
+sub copy_skel {
+    my ( $skel, $home ) = @_;
+
+    if ( $skel ne "" && %{ SCR->Read( ".target.stat", $skel ) } ) {
```
jreidinger, line 69: I would like to see here more logging as it can be tricky to debug issue here.

jreidinger (approved):

PR #242, opened 2021-04-01
Fix auto users reset

jreidinger:
bsc: https://bugzilla.suse.com/show_bug.cgi?id=1184216

Missing reset method result in calling abstract method that lead to exception.


lslezak (approved):

teclator:
:heavy_check_mark: Created OBS submit [request #240509](https://build.suse.de/request/show/240509) manually

PR #246, opened 2021-04-22
Adapt namespace

joseivanlopez:
## Problem

New user classes were created under *Y2User* namespace, but it should be *Y2Users*. Moreover, it was decided to re-organize readers and writers. Now they are placed under its own namespace instead of a generic *Readers* or *Writers* namespace. This would make easier to create auxiliary classes for a specific type of back-end, for example: *Y2Users::UsersSimple::Common*, *Y2Users::UsersSimple::Reader*, *Y2Users::UsersSimple::Writer*.

## Solution

* Namespace was renamed
* Readers were moved
* Minor fixes

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,14 +1,10 @@
 # use the shared YaST defaults
 inherit_from:
-  /usr/share/YaST2/data/devtools/data/rubocop_yast_style.yml
+  /usr/share/YaST2/data/devtools/data/rubocop-0.71.0_yast_style.yml
 
 # Don't enforce any particular name for block params
 SingleLineBlockParams:
   Enabled: false
-
-# Enforce if/unless at the end only for really short lines
-Style/IfUnlessModifier:
-  MaxLineLength: 60
 
 AllCops:
   Exclude:
--- a/src/lib/y2users/user.rb
+++ b/src/lib/y2users/user.rb
@@ -19,11 +19,11 @@
 
 require "yast2/execute"
 
-module Y2User
+module Y2Users
   # Representing user configuration on system in contenxt of given User Configuration
   # @note Immutable class.
   class User
-    # @return [Y2User::Configuration] reference to configuration in which it lives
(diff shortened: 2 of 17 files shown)


src/lib/y2users/user.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    # @return [Array<String>] Fields in GECOS entry
+    attr_reader :gecos
+
+    # @return [:local, :ldap, :unknown] where is user defined
+    attr_reader :source
+
+    # @see respective attributes for possible values
+    def initialize(configuration, name,
+      uid: nil, gid: nil, shell: nil, home: nil, gecos: [], source: :unknown)
```
dgdavid, line 53: We have too much params here, but I prefer the method definition in just one line.

Perphaps changing `uid: nil, gid: nil, shell: nil, home: nil, gecos: [], source: :unknown` by `attrs = {}` and then using `@gecos = attrs[:gecos] || []` could help.

ancorgs (approved):
LGTM

.rubocop.yml
```diff
@@ -1,15 +1,11 @@
 # use the shared YaST defaults
 inherit_from:
-  /usr/share/YaST2/data/devtools/data/rubocop_yast_style.yml
```
dgdavid, line 3: If you're going to fix the Rubocop problems in a different PR, maybe you can drop this commit from this PR.

dgdavid (approved):

.rubocop.yml
joseivanlopez: The rubocop config file was broken. I think it can still be here. Otherwise, all the real rubocop issues would not be exposed at all.

src/lib/y2users/user.rb
joseivanlopez: Yes, this is something to improve. Having named params are useful to avoid passing invalid params. If we use a hash instead of named parameters, we should also add control for invalid values in the hash. But yes, we have to do something.

PR #247, opened 2021-04-22
WIP: skeleton of the Y2Users::Linux writer

ancorgs:
We are developing a writer that commits the changes to the real system. This PR includes the skeleton (that is, the right files in the right places) with no real functionality. It also includes some documentation and details about the future implementation.


dgdavid (approved):
Looks good to keep working on top of it, thanks! 

So, please, merge it if you don't mind :)

PR #249, opened 2021-04-23
Create home directory and distinguish system users

ancorgs:
With this, the Linux writer always tries to create the home directory for normal users, but not for system ones. We agreed that should be the expected behavior.

This also improves a bit some of the existing tests of the Linux writer (for example, making sure the `chpasswd` command does not reveal too much).


dgdavid (approved):
LGTM

PR #250, opened 2021-04-23
Linux writer implementation

dgdavid:
This is a follow-up of #247. It starts adding functionality in Y2Users::Linux::Writer, making use of [`useradd`](https://man7.org/linux/man-pages/man8/useradd.8.htmlhttps://man7.org/linux/man-pages/man8/useradd.8.html) and [`chpasswd`](https://man7.org/linux/man-pages/man8/chpasswd.8.html) commands.

It adds basic unit tests too, but it's expected to improve them (see #249)


ancorgs (approved):
LGTM

PR #251, opened 2021-04-23
Add UsersSimple writer

joseivanlopez:
## Problem

During the installation, we plan to create new *Y2Users::User* objects as result of the users forms. But we need to save the users information into the *UsersSimple* module because several reason:

* As a way of saving the users data for recovering it later when the users are going to be created in the system.
* As a way of ensuring that other modules can still access to the users data from *UsersSimple*, as they could be doing now.

## Solution

Added new writer (*Y2Users::UsersSimple::Writer*) that stores information from *Y2Users::User* objects into the *UsersSimple* module.

NOTES:

* *Y2Users::Password* contains an encrypted password, but *UsersSimple* expects a plain password. Maybe,  *Y2Users::Password* should store a plain password too.
* *UsersSimple* has some separate attributes in order to save some specific data, for example the root password. Right now, those attributes are not populated by the writer. The root password is stored as part of the root user.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+# https://EditorConfig.org  -*- ini -*-
+# This is a unified way to tell all your editors
+# about the proper indentation style in this repo.
+#
+# Emacs:   https://github.com/editorconfig/editorconfig-emacs#readme
+# Vim:     https://github.com/editorconfig/editorconfig-vim#readme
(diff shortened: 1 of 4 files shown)


.editorconfig
```diff
@@ -0,0 +1,19 @@
+# https://EditorConfig.org  -*- ini -*-
```
dgdavid, line 1: Nice to see the .editorconfig around :)

dgdavid (approved):
LGTM

PR #253, opened 2021-04-26
UsersSimple Writer: manage root user info

joseivanlopez:
## Problem

*UsersSimple* module is not expected to include the *root* user in its list of users. But *Y2Users::UsersSimple::Writer* is not excluding the root user when writing users into *UsersSimple*.

## Solution

*Y2Users::UsersSimple::Writer* does not store *root* user anymore. But it uses the *UsersSimple* API to store root related info like the password.

### DIFF
--- a/src/lib/y2users/user.rb
+++ b/src/lib/y2users/user.rb
@@ -113,6 +113,13 @@ module Y2Users
       ATTRS.all? { |a| public_send(a) == other.public_send(a) }
     end
 
+    # Whether it is the root user
+    #
+    # @return [Boolean]
+    def root?
+      name == "root"
+    end
+
     # Whether this is a system user
     #
     # This is important for several reasons:
--- a/src/lib/y2users/users_simple/writer.rb
+++ b/src/lib/y2users/users_simple/writer.rb
@@ -23,7 +23,7 @@ Yast.import "UsersSimple"
 
 module Y2Users
   module UsersSimple
-    # Class for writing users configuration into the old UsersSimple Yast Module
+    # Class for writing users config into the old UsersSimple Yast Module
     class Writer
       # Constructor
       #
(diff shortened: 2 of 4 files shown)


src/lib/y2users/users_simple/writer.rb
```diff
@@ -33,9 +33,18 @@ def initialize(config)
...
+      # root_public_key).
       def write
-        users_simple = config.users.map { |u| to_user_simple(u) }
+        users_simple = config.users.reject(&:root?).map { |u| to_user_simple(u) }
         Yast::UsersSimple.SetUsers(users_simple)
+
+        root = config.users.find(&:root?)
+        return unless root
+
+        Yast::UsersSimple.SetRootPassword(root.password.value) unless root.password.nil?
```
jreidinger, line 47: please add todo as I think here we set plain password and not encrypted one, right?

src/lib/y2users/user.rb
```diff
@@ -113,6 +113,13 @@ def ==(other)
       ATTRS.all? { |a| public_send(a) == other.public_send(a) }
     end
 
+    # Whether it is the root user
+    #
+    # @return [Boolean]
+    def root?
+      name == "root"
```
jreidinger, line 120: that reminds me another trick to prevent ssh attacks...rename root to toor, but I expect it is really corner case and as we have this only for installation, it should not be issue.

src/lib/y2users/users_simple/writer.rb
```diff
@@ -23,7 +23,7 @@
 
 module Y2Users
   module UsersSimple
-    # Class for writing users configuration into the old UsersSimple Yast Module
+    # Class for writing users config into the old UsersSimple Yast Module
```
jreidinger, line 26: I am not sure if we should use acronym also in comment. That is why I keep in comments full name.

joseivanlopez: Ok, I can revert it. I replaced it automatically when fixing a conflict.

jreidinger (approved):

PR #254, opened 2021-04-26
nest password into user instead of having it in config

jreidinger:


src/lib/y2users/linux/reader.rb
```diff
@@ -96,25 +99,36 @@ def read_groups(config)
...
+        max_age = values[SHADOW_MAPPING["maximum_age"]]
+        inactivity_period = values[SHADOW_MAPPING["inactivity_period"]]
+        expiration = parse_account_expiration(values[SHADOW_MAPPING["account_expiration"]])
+        Password.new(
+          config,
+          values[SHADOW_MAPPING["username"]],
+          value:              values[SHADOW_MAPPING["value"]],
+          last_change:        parse_last_change(values[SHADOW_MAPPING["last_change"]]),
+          minimum_age:        values[SHADOW_MAPPING["minimum_age"]].to_i,
+          maximum_age:        max_age&.to_i,
```
dgdavid, line 125: NP: Why use `values[SHADOW_MAPPING["maximum_age"]]&.to_i` directly here too? The same for the rest.

dgdavid (approved):
LGTM, just a minor comment

src/lib/y2users/linux/reader.rb
jreidinger: well, reason is different meaning. Maximum age set to 0 means something different then if not set. In minimum age it is same ( disabled, in max age it should not be used as it age immediatelly )

dgdavid: Not sure if I got it. Probably I'm missing something, but at first glance both approaches looks the same for me.

I mean, 

```ruby 

max_age = values[SHADOW_MAPPING["maximum_age"]]
something = max_age&.to_i
```

will produce the same result than

```ruby
something = values[SHADOW_MAPPING["maximum_age"]]&.to_i
```

What am I missing?

jreidinger: yes, they are same. Just it start to be too long, so use separate value. No other reason.

dgdavid: Yes, they looks too long. However, my proposal was just to keep consistence and use the same way to _assign_ the value for all of them (_minimum\_age_ and _maximum\_age_ have equals length). I have no strong opinion about which choose, although I tend to avoid the intermediate variable :)

Anyway, still **LGTM** :stuck_out_tongue_winking_eye:

src/lib/y2users/user.rb
```diff
@@ -103,7 +101,10 @@ def full_name
     def clone_to(config)
       attrs = ATTRS.each_with_object({}) { |a, r| r[a] = public_send(a) }
       attrs.delete(:name) # name is separate argument
-      self.class.new(config, name, attrs)
+      cloned = self.class.new(config, name, attrs)
+      cloned.password = password.clone_to(config) if password
```
joseivanlopez, line 105: Should not it be `password.clone_to(user)`? I think we don't need to assign a password to a config anymore. Otherwise, we can end with passwords assigned to an user that does not belong to the config of the password.

dgdavid: :thinking: :thinking: Good catch!

jreidinger: well, it is for consistency. For each object it is important under which context it lives ( aka system, staging, etc ). Of course it can be user and then transitent config as user lives in that context, but I prefer consistency. It e.g. allows easier changes in some staging graf. If you modify user you simply create new one with new parameters and replace old one and just assign password to that user if it is same.
Situation that password belong to user with different config is of course issue, but it should be easily detected and is same issue as if user has group that does not exist in given config.

src/lib/y2users/config.rb
```diff
@@ -18,7 +18,7 @@
 # find current contact information at www.suse.com.
 
 module Y2Users
-  # Holds references to elements of user configuration like users, groups or passwords.
+  # Holds references to elements of user configuration like users or groups.
```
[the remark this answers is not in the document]
jreidinger: Sorry, I do not get this comment.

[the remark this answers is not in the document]
joseivanlopez: This class does `Password.new`. I think we need to *require "password"*.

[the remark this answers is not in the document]
jreidinger: I have to be blink, but I do not see any place where config initialize any Object beside self.

src/lib/y2users/user.rb
joseivanlopez: Yes, I see your point about having objects under a specific context. But for the password, I think the user is who defines the context. Maybe we can do something like:

~~~
cloned_user.password = password.duplicate
~~~

What do you think?

jreidinger: yes, as said on call it is possible to have password live just as reference in user and clone it with simple `cloned_user.password = password.clone`

src/lib/y2users/config.rb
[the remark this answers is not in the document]
joseivanlopez: Sorry, I mixed this file with the following one (reader.rb).

PR #255, opened 2021-04-26
[Y2Users::Linux::Writer] Creates only new users

dgdavid:
Following with  `Y2Users::Linux::Writer` implementation (see #247, #249, and #250), now it 

* receives the _initial configuration_, and
* only creates new users or modifies passwords of existing ones

### DIFF
--- a/src/lib/y2users/linux/writer.rb
+++ b/src/lib/y2users/linux/writer.rb
@@ -24,8 +24,9 @@ module Y2Users
     # Writes users and groups to the system using Yast2::Execute and standard
     # linux tools.
     #
-    # NOTE: currently it only creates new users, removing or modifying users is
-    # still not covered. No group management either.
+    # NOTE: currently it only creates new users or modifies passwords of
+    # existing ones. Removing or fully modifying users is still not covered.
+    # No group management either.
     #
     # A brief history of the differences with the Yast::Users (perl) module:
     #
@@ -90,7 +91,10 @@ module Y2Users
 
       # Performs the changes in the system
       def write
-        config.users.map { |user| add_user(user) }
+        config.users.map do |user|
+          add_user(user) if new_user?(user)
+          set_password(user)
+        end
         # TODO: update the NIS database (make -C /var/yp) if needed
         # TODO: remove the passwd cache for nscd (bug 24748, 41648)
         nil
@@ -126,15 +130,36 @@ module Y2Users
       CHPASSWD = "/usr/sbin/chpasswd".freeze
       private_constant :CHPASSWD
 
+      # Whether given user does not exist yet
+      #
+      # FIXME: choose a better criteria to identify a new user. An internal Y2User::User id?
+      #
+      # @param user [Y2User::User]
+      # @return [Boolean] true if there is a user with same name in @initial_config; false otherwise
+      def new_user?(user)
+        @initial_users_names ||= initial_config.users.map(&:name)
+
+        !@initial_users_names.include?(user.name)
+      end
+
+      # Executes the command for creating the user
+      #
+      # @param user [Y2User::User]
       def add_user(user)
         Yast::Execute.on_target!(USERADD, *useradd_options(user))
+      rescue Cheetah::ExecutionFailed => e
+        log.error("Error creating user '#{user.name}' - #{e.message}")
+      end
+
+      # Executes the command for setting the password of given user
+      #
+      # @param user [Y2User::User]
+      def set_password(user)
+        return unless user.password&.value
+
         Yast::Execute.on_target!(CHPASSWD, *chpasswd_options(user)) if user.password&.value
       rescue Cheetah::ExecutionFailed => e
-        if e.message.include?(USERADD)
-          log.error("Error creating user '#{user.name}' - #{e.message}")
-        else
-          log.error("Error setting password for '#{user.name}' - #{e.message}")
-        end
+        log.error("Error setting password for '#{user.name}' - #{e.message}")
       end
 
       # Generates and returns the options expected by `useradd` for given user
--- a/src/lib/y2users/linux/writer.rb
+++ b/src/lib/y2users/linux/writer.rb
@@ -99,16 +99,16 @@ module Y2Users
       #
       # @return [Y2Issues::List] the list of issues found while writing changes; empty when none
       def write
-        @issues = Y2Issues::List.new
+        issues = Y2Issues::List.new
 
         config.users.map do |user|
-          add_user(user)
-          change_password(user)
+          add_user(user, issues)
+          change_password(user, issues)
         end
         # TODO: update the NIS database (make -C /var/yp) if needed
         # TODO: remove the passwd cache for nscd (bug 24748, 41648)
 
-        @issues
+        issues
       end
 
     private
@@ -123,11 +123,6 @@ module Y2Users
       #
       # @return [Y2User::Config]
       attr_reader :initial_config
-
-      # The list of issues generated while writting changes to the system. See #write
-      #
-      # @return [Y2Issues::List]
-      attr_reader :issues
 
       # Command for creating new users
       USERADD = "/usr/sbin/useradd".freeze
@@ -160,8 +155,9 @@ module Y2Users
 
       # Executes the command for creating the user
       #
-      # @param user [Y2User::User]
-      def add_user(user)
+      # @param user [Y2User::User] the user to be created on the system
+      # @param issues [Y2Issues::List] a collection for adding an issue if something goes wrong
+      def add_user(user, issues)
         return unless new_user?(user)
 
         Yast::Execute.on_target!(USERADD, *useradd_options(user))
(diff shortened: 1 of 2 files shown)


src/lib/y2users/linux/writer.rb
```diff
@@ -124,15 +130,38 @@ def write
       CHPASSWD = "/usr/sbin/chpasswd".freeze
       private_constant :CHPASSWD
 
+      # Whether given user does not exist yet
+      #
+      # FIXME: choose a better criteria to identify a new user. An internal Y2User::User id?
```
jreidinger, line 135: I think internal ID is not good as it can have problem with some operations. Here the biggest topic is to distinguish between two cases:

1. User is deleted and new one is created with same ID -> new home, so old user data is removed
2. rename of user so same ID -> keep old home, do not remove any data!

change uid of user is different topic and very tricky as it means to traverse whole disk and chown from old one to new one...probably we should not support it.

src/lib/y2users/linux/writer.rb
```diff
@@ -124,15 +130,38 @@ def write
...
+
+      # Executes the command for creating the user
+      #
+      # @param user [Y2User::User]
       def add_user(user)
+        return unless new_user?(user)
+
         Yast::Execute.on_target!(USERADD, *useradd_options(user))
+      rescue Cheetah::ExecutionFailed => e
+        log.error("Error creating user '#{user.name}' - #{e.message}")
```
jreidinger, line 153: is log enough? what about using at least that Issues List @imobachgs adds to yast2?

src/lib/y2users/linux/writer.rb
```diff
@@ -124,15 +130,38 @@ def write
...
+      end
+
+      # Executes the command for setting the password of given user
+      #
+      # @param user [Y2User::User]
+      def change_password(user)
+        return unless user.password&.value
+
```
jreidinger, line 161: I see here big three TODOs that I would like to have mentioned in code ( I know none of it is needed for installation ):

1. user modification
2. other password configuration
3. user remove

src/lib/y2users/linux/writer.rb
```diff
@@ -24,8 +24,9 @@ module Linux
     # Writes users and groups to the system using Yast2::Execute and standard
     # linux tools.
     #
-    # NOTE: currently it only creates new users, removing or modifying users is
-    # still not covered. No group management either.
+    # NOTE: currently it only creates new users or modifies passwords of
+    # existing ones. Removing or fully modifying users is still not covered.
+    # No group management either.
```
jreidinger, line 29: also no group configuration

src/lib/y2users/linux/writer.rb
```diff
@@ -107,6 +107,12 @@ def write
       # @return [Y2User::Config]
       attr_reader :config
 
+      # Initial state of the system that will be compared with {#config} to know what changes need
+      # to be performed
```
jreidinger, line 111: I would add here note that this is typically `Y2User::Config.system`

src/lib/y2users/linux/writer.rb
```diff
@@ -82,19 +82,19 @@ class Writer
...
-      # NOTE: right now we consider the system is empty, so we only receive one parameter.
-      #   But in the future we might need another configuration describing the initial state,
-      #   so we can compare and know what changes are actually needed.
-      #
-      # @param config [Y2User::Config] configuration containing the users
-      #   and groups that should exist in the system after writing
-      def initialize(config)
+      # @param config [Y2User::Config] see #config
+      # @param initial_config [Y2User::Config] see #initial_config
+      def initialize(config, initial_config)
```
jreidinger, line 87: Maybe we can use system as default here so 

```suggestion
      def initialize(config, initial_config = Y2User::Config.system)
```

test/lib/y2users/linux/writer_test.rb
```diff
@@ -100,6 +103,21 @@
       allow(Yast::Execute).to receive(:on_target!)
     end
 
+    context "for an existing user" do
+      before do
+        initial_config.users << user
+        initial_config.passwords << password
```
jreidinger, line 109: this collide with https://github.com/yast/yast-users/pull/254. We would need to update one or another PR. ( just note )

src/lib/y2users/linux/writer.rb
dgdavid: I was discussing it with @ancorgs last week: looks like a previous job for the validator.

src/lib/y2users/linux/writer.rb
dgdavid: They are somehow stated few lines above, on the user class documentation. Anyway, I'll recheck them tomorrow. Thanks!

src/lib/y2users/linux/writer.rb
```diff
@@ -124,15 +130,38 @@ def write
...
+      # Whether given user does not exist yet
+      #
+      # FIXME: choose a better criteria to identify a new user. An internal Y2User::User id?
+      #
+      # @param user [Y2User::User]
+      # @return [Boolean] true if there is a user with same name in @initial_config; false otherwise
+      def new_user?(user)
+        @initial_users_names ||= initial_config.users.map(&:name)
+
+        !@initial_users_names.include?(user.name)
```
dgdavid, line 142: Self note: this check should be reviewed and, most probably, reviewed. Looks like the username can be changed too.

src/lib/y2users/linux/writer.rb
jreidinger: Well, my POV is that if it should not happen as validator already catch it, then raise exception as we have bug in validator....and if it can happen, we need to report it. Silently ignore it is the worst case as it means that user do not know about failure and result is not what user expect.

dgdavid: Yes, you're right. What I meant was not to perform validations _again_, but it's true that we should provide some kind of information to the user at some point :) But I was thinking to do it later, in another step/PR

src/lib/y2users/linux/writer.rb
dgdavid: I discussed it again with @ancorgs yesterday and the conclusion was to not have a default value (yet) since the initial_config might be different depending in the mode the code is being executed (installation, running system, auto upgrade, auto install, firstboot...).


Let's keep it as it is by now.

src/lib/y2users/linux/writer.rb
dgdavid: See https://github.com/yast/yast-users/issues/257

src/lib/y2users/linux/writer.rb
dgdavid: Creating and returning the list of found issues. See https://github.com/yast/yast-users/pull/255/commits/084fe46c45066171e1870c5efd978fa8a52e237f

It's simple / dummy by now on purpose. Let's see how evolve it while the writer grow.

ancorgs (approved):
Looks good enough for the time being. Of course we will have to refine it very soon (eg. when merging the new API with internal ids), but looks like a solid start worth merging.

src/lib/y2users/linux/writer.rb
```diff
@@ -79,25 +82,31 @@ module Linux
...
+
+        config.users.map do |user|
+          add_user(user)
+          change_password(user)
+        end
         # TODO: update the NIS database (make -C /var/yp) if needed
         # TODO: remove the passwd cache for nscd (bug 24748, 41648)
-        nil
+
+        issues
```
jreidinger, line 109: please document return value in yardoc

src/lib/y2users/linux/writer.rb
```diff
@@ -107,6 +116,17 @@ def write
...
+      # Initial state of the system (usually a Y2User::Config.system in a running system) that will
+      # be compared with {#config} to know what changes need to be performed.
+      #
+      # @return [Y2User::Config]
+      attr_reader :initial_config
+
+      # The list of issues generated while writting changes to the system
+      #
+      # @return [Y2Issues::List]
+      attr_accessor :issues
```
jreidinger, line 128: is accessor needed when you return it in write? it makes method hardly re-entrant. If there is reason, it is fine for me.

dgdavid: Nope, there is no a reason apart from a _leftover_ from my side :)

src/lib/y2users/linux/writer.rb
```diff
@@ -124,10 +124,10 @@ def write
       # @return [Y2User::Config]
       attr_reader :initial_config
 
-      # The list of issues generated while writting changes to the system
+      # The list of issues generated while writting changes to the system. See #write
       #
       # @return [Y2Issues::List]
-      attr_accessor :issues
+      attr_reader :issues
```
jreidinger, line 130: I am still not sure why it is needed. Do you have example of usage? I kind of expect something like
```ruby
issues = Linux::Writer.write
if !issues.empty?
  ... reporting
end
```

dgdavid: It's a private `attr_reader` just for _documenting_ it and to avoid using `@` all the time (now `issues << Y2Issues::Issue.new(...)` is used in two different methods.

I can drop it if it hurts :)

jreidinger: Well, what I mean that I would not use `@issues` at all and just pass it in given methods. This way you do not hold issues from previous runs.

dgdavid: Ah, I see what you meant now

jreidinger (approved):

PR #256, opened 2021-04-27
Add plain password field and protect attributes

jreidinger:


src/lib/y2users/password.rb
```diff
@@ -18,20 +18,29 @@
...
     #   - "!" or "*" is disabled login by password
     #   - "" password-less login allowed
     #   - "!..." disabled password. After exclamation mark is old value that no longer can be used
     #     for login
     #   - nil means password is not yet set
-    attr_reader :value
+    secret_attr :value
+
+    # @!attribute plain_value
+    # @return [String, nil] Plain unexcrypted password. Nil means not set or
```
ancorgs, line 41: typo: unexcrypted

src/lib/y2users/password.rb
```diff
@@ -18,20 +18,29 @@
 # find current contact information at www.suse.com.
 
 require "yast2/execute"
+require "yast2/secret_attributes"
```
joseivanlopez, line 21: Please, adapt the spec file in order to depend on the new *yast2* that provides *secret_attributes*.

src/lib/y2users/password.rb
```diff
@@ -65,7 +74,7 @@ def initialize(config, name, value: nil, last_change: nil, minimum_age: nil,
       account_expiration: nil, source: :unknown)
```
ancorgs, line 74: I know we already have tons of keyword arguments here, but having `value` and not `plain_value` looks wrong. In almost all cases you will initialize the object with one of those two arguments (all the others are less common).

src/lib/y2users/password.rb
```diff
@@ -18,20 +18,29 @@
...
 
+    # @!attribute value
     # @return [String, nil] Encrypted password. It can have several specific values:
     #   - "!" or "*" is disabled login by password
     #   - "" password-less login allowed
     #   - "!..." disabled password. After exclamation mark is old value that no longer can be used
     #     for login
     #   - nil means password is not yet set
-    attr_reader :value
+    secret_attr :value
```
ancorgs, line 38: nil means password is not yet set in its final form, it still may be set as `plain_value`

src/lib/y2users/users_simple/writer.rb
```diff
@@ -44,8 +44,7 @@ def write
         root = config.users.find(&:root?)
         return unless root
 
-        # TODO: save plain password instead of encrypted one.
-        Yast::UsersSimple.SetRootPassword(root.password.value) unless root.password.nil?
+        Yast::UsersSimple.SetRootPassword(root.password.plain_value) unless root.password.nil?
```
joseivanlopez, line 47: Please, see the next TODO at line 69. It should be adapted too.

src/lib/y2users/password.rb
ancorgs: Moreover, it would maybe be useful to have some method like `#any_value?` to check if either of both is set.

src/lib/y2users/users_simple/reader.rb
```diff
@@ -38,8 +38,19 @@ def read_to(config)
...
           user.password = Password.new(config, user["uid"],
             value: Yast::Builtins.cryptsha512(user["userPassword"]))
+          user.password.plain_value = user["userPassword"]
           config.users << user
         end
+
+        # Read also root user settings
+        root_pwd_plain = Yast::UsersSimple.GetRootPassword
+        root_pwd = Password.new(config, "root",
+          value: Yast::Builtins.cryptsha512(user["userPassword"]))
```
ancorgs, line 48: If, at this point, we only know the plain password I would only set `plain_value`. In my opinion, the transformation of that into a final `value` is not responsibility of this reader.

ancorgs: This introduces a risk of getting both passwords out of sync if something like this is executed (creative usage of the API at some points for clarity):

```ruby
initial = Y2Users::Config.new(:initial)
# The following call should only set #plain_password, otherwise...
Y2Users::UsersSimple::Reader.new.read_to(initial)

config2 = initial.clone
# Now we introduce inconsistencies between both attributes
config2.root_user.password.plain_value = "S3cr3t"

# This should be the point in which #value should be calculated based on #plain_password
# it may even use a different crypt algorithm
Y2Users::Whatever::Writer.new(config2).write
```

ancorgs: In that regard, I would even consider that maybe the `Password#plain_value=` setter should set `#value` to nil.

ancorgs: Thinking about the problem of both values going out of sync... maybe we should have just one value and a separate attribute to indicate whether such value is plain or already encrypted. Something like this (just a first wild idea):

```ruby
class Password
  attr_reader :value

  def initialize(value, plain: true)
    @value = value
    @plain_value = plain
  end

  def plain?
    !!@plain_value
   end

  def plain_value
    return nil unless plain?
    value
  end

  def encrypted_value
    return nil if plain?
    value
   end
end
```

jreidinger: It makes sense for me to have just one. At first I consider if we can do it without flag, but it is a bit problematic as value is even more tricky as it can be even something like disable login or even more tricky value like "!<value>" which is locked password with old value after ! which can be used to unlock account. So I am not sure how to show this domain difficulty into API. As we have 

- plain value
- encrypted value
- locked encrypted value
- disabled login ( by now represented in that encrypted value and sadly it is  "!" or "*" which can be valid plain passwords. )

I am even not sure in your snippet if we should keep that value reader which is context dependent.
Maybe we should instead use something like Enum in value? And that object answer its kind from kinds above and provide value when it makes sense.

So what about something like
```
class PasswordValue
  secret_attribute :value

  def initialize(value)
      @value = value
  end

  def encrypted?
     false
  end

  def plain?
      false
  end

  def locked?
      false
  end

  def disabled?
      false
  end
end

class EncryptedPassword < PasswordValue
   def initialize(value)
     super
   end

  def encrypted?
      true
  end
end

class PlainPassword
   def initialize(value)
     super
   end

  def plain?
      true
  end
end

...
```

src/lib/y2users/users_simple/writer.rb
jreidinger: good catch

src/lib/y2users/users_simple/reader.rb
ancorgs: As I understand this, the idea is to then also have subclasses for `LockedPassword` and `DisabledPassword`. If that's the case, I see some overlapping between the `PasswordValue#value` and the class hierarchy. For example, what if I do this?

```ruby
example = EncryptedPassword.new("*")
```

Moreover, if I want to create a `DisabledPassword` object, which `value` should I pass to the constructor?

Or maybe I understood it wrong and the idea is to only have `EncryptedPassword` and `PlainPassword`, with an implementation of `EncryptedPassword` similar to this:

```ruby
class EncryptedPassword < PasswordValue
   def initialize(value)
     super
   end

  def encrypted?
      true
  end

  def locked?
    value && value.start_with("!")
  end

  def disabled?
    value == "*"
  end
end
```

If that's the case, I like it more. But I'm not sure if having the `Y2Users::Password` and then a `Y2Users::PasswordValue` may be a bit over the top.

jreidinger: > As I understand this, the idea is to then also have subclasses for `LockedPassword` and `DisabledPassword`. If that's the case, I see some overlapping between the `PasswordValue#value` and the class hierarchy. For example, what if I do this?
> 
> ```ruby
> example = EncryptedPassword.new("*")
> ```

this should raise as it is invalid encrypted value

> 
> Moreover, if I want to create a `DisabledPassword` object, which `value` should I pass to the constructor?

nothing, it should have empty constructor and pass that "*" inside it like
```ruby
def initialize
  super("*")
end
```

> 
> Or maybe I understood it wrong and the idea is to only have `EncryptedPassword` and `PlainPassword`, with an implementation of `EncryptedPassword` similar to this:
> 
> ```ruby
> class EncryptedPassword < PasswordValue
>    def initialize(value)
>      super
>    end
> 
>   def encrypted?
>       true
>   end
> 
>   def locked?
>     value && value.start_with("!")
>   end
> 
>   def disabled?
>     value == "*"
>   end
> end
> ```
> 
> If that's the case, I like it more. But I'm not sure if having the `Y2Users::Password` and then a `Y2Users::PasswordValue` may be a bit over the top.

Yeah, I do not like naming...But I am not sure if we should rename Y2Users::Password to Y2Users::Shadow and let live just PasswordValue? BTW if it helps, we can also have that subclasses as nested ones and have helper constructors in PasswordValue like
```ruby
class PasswordValue
   def self.create_encrypted(value)
      PasswordValue::EncryptedPassword.new(value)
   end
end
```

joseivanlopez: IMHO, something like what @ancorgs proposes should be enough, shouldn't it?

~~~
class Password
  secret_attribute :value

  def initialize(value, plain: true)
    self.value = value
    @plain = plain
  end

  def plain?
    !!@plain
  end

  def encrypted?
    !plain?
  end

  def locked?
    value.start_with("!")
  end

  def disabled?
    value == "*"
  end
end
~~~

And maybe, I would not allow to modify a password. We can create a new one if we need a new password. Otherwise, we could still end with wrongly configured passwords:

~~~
pwd = Password.new("S3cr3t")
pwd.value = encrypt("My new password")
pwd.plain? #=> true
~~~

jreidinger: What I do not like about @ancorgs proposal is that it is hard to keep it consistent as you see with setting value. In fact even your example has bug as if value is "!" it is disabled and not locked. Locked is only if it start with "!" but follows with old encrypted code. Also it lets a lot of responsibility to Password configuration, but it is all related password value type. So instead of trying to make value and all flags consistent, own object is always consistent as it prevent to have mixture of different value types.

jreidinger: BTW if I take that 
```
pwd.change("!1afads34fes", :encrypted)
pwd.encypted? #=> true
pwd.locked? #=> true
```
and instead use object it is more readable for me:
```
pwd.value = Password.encrypted_value("!1afads34fes")
pwd.value.encypted? #=> true
pwd.value.locked? #=> true
```

src/lib/y2users/password.rb
```diff
@@ -18,20 +18,20 @@
...
-    #   - "!" or "*" is disabled login by password
-    #   - "" password-less login allowed
-    #   - "!..." disabled password. After exclamation mark is old value that no longer can be used
-    #     for login
-    #   - nil means password is not yet set
-    attr_reader :value
+    # @!attribute value
+    # @return [Value, nil] password value. Can be any subclass of Value. nil when password is
+    #   not set at all.
+    secret_attr :value
```
ancorgs, line 34: Does this need to be a `secret_attr`? I think that making `PasswordValue::content` secret is enough. I suspect this adds nothing but noise.

src/lib/y2users/password.rb
```diff
@@ -93,5 +103,46 @@ def ==(other)
       # do not compare configuration to allow comparison between different configs
       ATTRS.all? { |a| public_send(a) == other.public_send(a) }
     end
+
+    # Represents password value. It specific type is defined as subclass and can be queried
```
joseivanlopez, line 107: Typo: Its specific type ...

src/lib/y2users/password.rb
```diff
@@ -93,5 +103,46 @@ def ==(other)
...
+    class Value
```
ancorgs, line 108: I don't like to have nested classes when we intent to use them from the outside world.

In other words, having a class `Password` and then a class `Password::Value` is fine if the inner one is only known to `Password`. But since we have a `Password#value` method that returns an object of the `Value` class, I would rather have `Password` and `PasswordValue`.

At least, that's the criterion we have followed in other parts like storage-ng.

src/lib/y2users/password.rb
```diff
@@ -93,5 +103,46 @@ def ==(other)
...
+      def plain?
+        false
+      end
+
+      def encrypted?
+        false
+      end
+    end
+
+    # Represents encrypted password value or its specific types like disabled of locked logging.
```
joseivanlopez, line 126: s/of/or

src/lib/y2users/password.rb
```diff
@@ -57,6 +57,16 @@ class Password
     # @return [:local, :ldap, :unknown] where is user defined
     attr_reader :source
 
+    def self.plain(value)
```
joseivanlopez, line 60: NP: Maybe *new_plain* would be more auto-explanatory?

src/lib/y2users/password.rb
joseivanlopez: I have no strong opinion about this. In general, I like nested class names when it makes sense, even to use them publicly. But if we are already following a criterion, then I would respect it as much as possible.

dgdavid: I have no strong opinion for this concrete case either, but I tend to avoid nesting. So, I prefer `PasswordValue` over `Password::Value`.

src/lib/y2users/password.rb
```diff
@@ -93,5 +103,46 @@ def ==(other)
...
+      def encrypted?
+        true
+      end
+
+      def locked?
+        content.start_with?("!$")
+      end
+
+      def disabled?
+        ["*", "!"].include?(content)
```
dgdavid, line 137: Not sure if this is because 

> If the password field contains some string that is not a valid result of crypt(3), for instance ! or *, the user will not be able to use a unix password to log in (but the user may log in the system by other means). — _[man shadow(5)](https://linux.die.net/man/5/shadow)_

If so, there are other characters that should be included, 

>  It will be entirely printable ASCII, and will not contain **whitespace** or the characters ‘**:**’, ‘**;**’, ‘**\***’, ‘**!**’, or ‘**\\**’. — _[man crypt(3)](https://linux.die.net/man/3/crypt)_

dgdavid: :thinking: :thinking: :thinking: 

I see now that you're honoring what is stated in the class [documentation](https://github.com/yast/yast-users/blob/9be29d36f9d6619082856f8942c9abd1b279b5de/src/lib/y2users/password.rb#L29-L32)

I guess it's a _common_ practice, since I was not able to found an official documentation about it beyond the people using an invalid character  in at the beginning, but not replacing it completely.

Moreover, `usermod`and `passwd` prefix the value with an exclamation mark to lock the password, which is checked by the `lock?` method.

src/lib/y2users/password.rb
jreidinger: To be honest I am not sure. Question is if it is secret if value is nil?

src/lib/y2users/password.rb
jreidinger: well, I do not like much that new prefix. I more prefer `create_plain` or `build_plain`, but both is quite long. But if you prefer more auto explanatory, it is fine for me.

src/lib/y2users/password.rb
jreidinger: Me neither and what is worry that it often depends on implementation. Encrypted password start with "$<hash encoded type like 4,5,x or 2y>$" so anything else beside "!<encrypted password>" is probably disabled. For now  I think that * and ! covers majority of usage at least on suse.

src/lib/y2users/password.rb
joseivanlopez: What I prefer about nested names is that `Password::Value` states that `Value` is something that only makes sense in the context of `Password`, I mean, it is something to be used for a password and nothing else. Otherwise, `PasswordValue` is only a name, and nothing indicates that it is strongly attached to a `Password`. But again, a matter of taste.

ancorgs: Just to clarify, I don't have such a strong opinion either and I also see advantages in the nested approach. But I remember that we agreed in the rule of "nested only if not visible from the outside" back then in the context of storage-ng and I think it was intended as a general rule for all the new naming. But I can be wrong and maybe the rule does not apply here.

@imobachgs Do you remember better?

src/lib/y2users/password.rb
```diff
@@ -18,20 +18,19 @@
 # find current contact information at www.suse.com.
 
 require "yast2/execute"
+require "yast2/secret_attributes"
 
 module Y2Users
   # Password configuration for user including its hashed value.
   class Password
+    include Yast2::SecretAttributes
```
ancorgs, line 26: This is not longer needed here.

src/lib/y2users/password.rb
imobachgs: From https://gist.github.com/imobachgs/d90940315bd0471ee00b2a68ba22fbe5#filesnamespaces-organization I can read that nested classes are:

* Only for private APIs.
* Avoid cases like the storage Proposal code when possible.

I do not have a strong opinion. I tend to avoid nested classes but just because I do not want to end up with a 2000 LoC file containing a lot of stuff. However, I am fine with keeping ancillary (*and simple*) classes nested.

About the `Password::Value` vs `PasswordValue` discussion, it is secondary to me. I am sure that we can find many classes whose instances are only relevant when combined with another specific class, and they are not nested.

Finally, if the `Password::Value` class is simple enough and we do not plan to extend it in the future, just keep it nested. Otherwise, I would go for a separate one.

dgdavid: > Otherwise, `PasswordValue` is only a name, and nothing indicates that it is strongly attached to a `Password`

That's true

src/lib/y2users/password.rb
```diff
@@ -18,20 +18,18 @@
 # find current contact information at www.suse.com.
 
 require "yast2/execute"
+require "yast2/secret_attributes"
 
 module Y2Users
   # Password configuration for user including its hashed value.
   class Password
+
```
ancorgs, line 26: I would expect Rubocop to complain about this blank line

ancorgs (approved):
LGTM. After fixing rubocop issue.

Issue #257, opened 2021-04-27
API changes


joseivanlopez:
Current API for users/groups was designed with the idea of creating immutable objects. Finally we have decided to leave that idea and to go with a more "classical" API.

This issue summarizes a proposal for adapting the current API, which:

* Allows to create users/groups without an associated config (useful in some scenarios, e.g., testing).
* Explicitly registers/unregisters users/groups into a config.
* Does not allow to edit the list of users/groups directly (register/unregister instead).
* Password is directly assigned to an user.
* An user/group can be cloned, which would generate a new unregistered user/group. 

Here is how the new API could look:

~~~
john = User.new("john")
john.gid = 100
john.primary_group #=> nil
john.groups #=> []
john.password = Password.new(plain: "S3Cr3T")

john.attached? #=> false
john.id #=> nil
john.config #=> nil

group1 = Group.new("john", gid: 100)
group2 = Group.new("users", users_name: ["john"])

john.groups #=> []

config = Config.new
config.attach(john, group1, group2)

john.attached? #=> true
john.id #=> 1
john.config #=> config
john.primary_group #=> group1
john.groups #=> [group1, group2]

config2 = config.clone #=> clone users and groups, keeping the same internal ids.
john2 = config2.users.find { |u| u.name == john.name }
john2.gecos = ["New name"]
john2.is?(john) #=> true (same #id)
john2 == john #=> false (different attributes, e.g. #gecos)
john2.password == john.password #=> true

john3 = john.clone #=> clones the user without copying its id and config
john3.id #=> nil
john3.config #=> nil
jonh3.attach_to(config2)
john3.is?(john) #=> false
john3.password == john.password #=> true

config2.users #=> [john2, john3]
config2.detach(john3) # or john3.detach
john3.id #=> nil
john3.config #=> nil
config2.users #=> [john2]
~~~

jreidinger:
just few small notes to proposal:

- I prefer to have attach/detach instead of register/unregister. It sounds more familiar for me.
- is id writable?
- I am not sure if it is intention, but in example is used `user.group` but I still prefer `user.primary_group` to avoid confusion with groups method.
- not sure what will happen with if change user name with password, as passwords then cannot be same ( password contain user name )

ancorgs:
`id` is only written when attaching (and set to nil when detached). It cannot be changed via a normal accesor at any other point in time.

The `user.group` in the example is not an intentional change. It would still be `user.primary_group`.

[...]

PR #258, opened 2021-04-28
API changes

joseivanlopez:
## Problem

Initial API was designed with immutability concept in mind. Then, it was decided to follow a more classical approach,  see https://github.com/yast/yast-users/issues/257.

## Solution

API was modified as previously agreed.

NOTE: unit tests for `Config`, `User`, `Group` and `Password` will come with a separate PR. And the same for adapting the current readers/writers unit tests.

### DIFF
--- a/src/lib/y2users/config.rb
+++ b/src/lib/y2users/config.rb
@@ -18,9 +18,22 @@
 # find current contact information at www.suse.com.
 
 module Y2Users
-  # Holds references to elements of user configuration like users or groups.
-  # Class itself holds references to different configuration instances.
-  # TODO: write example
+  # Class to represent a configuration of users and groups
+  #
+  # @example
+  #   user = User.new("john")
+  #   group = Group.new("users")
+  #
+  #   config1 = Config.new("config1")
+  #   config1.users #=> []
+  #   config1.attach(user, group)
+  #   config1.users #=> [user]
+  #   config1.groups #=> [group]
+  #
+  #   config2 = config1.clone_as("config2")
+  #   user2 = config2.users.first
+  #   config2.detach(user2)
+  #   config2.users #=> []
   class Config
     class << self
       def get(name)
@@ -48,7 +61,6 @@ module Y2Users
           reader = Linux::Reader.new
         end
 
-        # TODO: make system config immutable, so it cannot be modified directly
         res = new(:system)
         reader.read_to(res)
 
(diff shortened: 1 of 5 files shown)


src/lib/y2users/config.rb
```diff
@@ -57,22 +56,90 @@ def system(reader: nil, force_read: false)
...
+        element.assign_config(config)
+        element.assign_internal_id(next_id) if element.id.nil?
+
+        @elements << element
+      end
+
+    private
+
+      def next_id
+        @next_id ||= max_id(elements)
```
jreidinger, line 136: I worry this is not much robust. E.g. consider that you have staging with newly attached user1 and then you have some dialog config that is cloned from system and it adds user2, then both get same id. I think we need global register of ids, so we ensure that it cannot accidentally be same.

joseivanlopez: Right now, cloning always generates a new config:

~~~
config2 = config1.clone_as("staging")
~~~

and you should only compare configs that were generated by cloning. Otherwise, ids from a config would collide with ids from another (not cloned) config.

ancorgs: I don't see a problem with that. If I got it correctly, the situation described by @jreidinger  is this:

- You start with a config called `system` with users read from the system.
- Then you clone it into another called `staging` and you add user1 there.
- Then you clone `system` again into another config called `another` and add user2 there.

So far, so good. What's the problem in having the same id in user1 and user2? I don't see anything wrong with that.

In the end you will do something like `System::Writer.new.write(staging, initial: system)` or maybe `System::Writer.new.write(another, initial: system)`. But for sure you will not do something like `System::Writer.new.write(another, initial: staging)` because it simply makes no sense.

jreidinger: well, my concern is that I expect that id is unique. So I can trust that if I see in logs that I have user with id X, it will be that user in whole logs. Now I need to find in logs ID and to which config it belongs and if we teoretically chain configs e.g. in wizard with config_step1, config_step2... and having also staging that is there for abort, then having mixture of ids is kind of critical for me. I just said that if we simply have global counter that just increase, we can be sure that id is always unique and probably I would also prefer to have users and groups unique id, so it is will be again easier to grep in logs.

ancorgs: I see @jreidinger's point about potentially confusing ids, although that has never been a problem so far in storage-ng, for example.

Anyway, I just wanted to point one thing. Chaining steps will still not introduce collision. config_step1 will be based on staging and config_step2 will be based on config_step1. So no collision there.

ancorgs: I believe the current algorithm has a problem. Steps to reproduce:

- Start with an empty config.
- Create a first user, `@next_id` is initialized to 0 and then increased by 1.
- Create two more, `@next_id` is increased so it now contains a 3. The user ids are 1, 2 and 3.
- Clone that config into another one called `staging`.
- Detach the user with id 3 from `staging`
- Create a new user in `staging`. `@next_id` is initialized to 2 and then increased by 1.
- Result: `staging` contains a user with id 3 that is not the same than the corresponding one in the initial config.

jreidinger: @ancorgs I think it is not reproducable as next_id is initialize once in staging to 3 and won't change later. But it can happen in that staging, step1 and step2. step1 detach user, step2 attach user and now staging and step2 contain user with same id.

src/lib/y2users/config.rb
```diff
@@ -48,31 +61,171 @@ def system(reader: nil, force_read: false)
...
+        @elements = []
+      end
+
+      # Attaches the element to the config
+      #
+      # @raise [RuntimeError] if the element is already attached
+      #
+      # @param element [User, Group]
+      def attach(element)
+        raise "Element already attached: #{element}" if element.attached?
```
dgdavid, line 210: Suggestion: since the element can be attached to other config too, perhaps worth it to include _where_ (the config name?) is attached.

src/lib/y2users/user.rb
```diff
@@ -17,70 +17,103 @@
...
     def primary_group
+      return nil unless attached?
+
       config.groups.find { |g| g.gid == gid }
     end
 
-    # @return [Array<Y2Users::Group>] list of groups where is user included including primary group
+    # Groups where the user is included. It also contains the primary group.
+    #
+    # The user must to be attached to a config in order to find its groups.
```
dgdavid, line 111: NP: s/must to be/must be

src/lib/y2users/user.rb
```diff
@@ -17,70 +17,103 @@
...
       # See #system?
       @system = false
     end
-    # rubocop: enable Metrics/ParameterLists
 
-    # @return [Y2Users::Group, nil] primary group set to given user or
-    #   nil if group is not set yet
+    # Primary group for the user
+    #
+    # The user must to be attached to a config in order to find its primary group.
```
dgdavid, line 100: NP: s/must to be/must be

src/lib/y2users/group.rb
```diff
@@ -17,57 +17,80 @@
...
-      @source = source
+      @users_name = []
+      @source = :unknown
     end
 
-    # @return [Array<Y2Users::User>] all users in this group, including ones that
-    # has it as primary group
+    # Users that become to this group, including users which have this group as primary group
+    #
+    # The group must to be attached to a config in order to find its users.
```
dgdavid, line 73: NP: s/must to be/must be

src/lib/y2users/user.rb
```diff
@@ -89,32 +122,29 @@ def expire_date
...
-    end
-
-    # Compares user object if all attributes are same excluding configuration reference.
-    # @return [Boolean] true if it is equal
+    # Whether two users are equal
+    #
+    # Only relevant attributes are compared. For example, the config in which the user is attached
+    # and the internal user id are not considered.
+    #
+    # @return [Boolean]
```
dgdavid, line 140: Note: so, this method cannot be used to infer _new_ users given two configurations, the "origin" and the "cloned". Instead, the `#is?` method from `ConfigElement` should be used. Right?

dgdavid (approved):
In general LGTM. I just left some minor comments there.

src/lib/y2users/config.rb
joseivanlopez: I expect that is shown as part of `element.inspect`.

dgdavid: Ah, right :sweat_smile:

src/lib/y2users/config.rb
```diff
@@ -18,9 +18,22 @@
...
-  # TODO: write example
+  # Class to represent a configuration of users and groups
+  #
+  # @example
+  #   user = User.new("john")
+  #   group = Group.new("users")
+  #
+  #   config1 = Config.new("config1")
+  #   config1.users #=> []
+  #   config1.attach(user, group)
```
jreidinger, line 29: I am not sure if this does not create impression that you need to attach user and group together

src/lib/y2users/user.rb
joseivanlopez: The only way to be sure that two users are actually the same is by means of the id. We cannot rely on the rest of attributes. So, `#is?` should be used to find new users, and `#==` to find changes between two users. Anyway, facility methods can be added to the API in order to compare configs.

src/lib/y2users/config.rb
joseivanlopez: Ok, I will modify the example.

PR #259, opened 2021-04-28
Adapt readers and writers to use the new API

dgdavid:
## Problem

Readers and writers for `Y2Users::Linux` and `Y2Users::UsersSimple` got updated after migrating to the new API.

## Solution

Adapt readers and writers to the new API.


jreidinger:
note: I found that there is slight overlap with https://github.com/yast/yast-users/pull/264 where I need to adapt linux/reader 
Do you plan to write test cases for reader or should I do it there?

ancorgs (approved):
LGTM

PR #260, opened 2021-04-29
move register responsibility to config manager

jreidinger:
Ideas:

- keep register simple hash where you can store any id associated with config
- config itself does not need name
- keep only specific one that is system
- as clone no longer need argument use plain ruby clone method


src/lib/y2users/config_manager.rb
```diff
@@ -0,0 +1,48 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Y2Users
+  # Holds references to different configuration instances.
+  # It is not mandatory for config to be registered here.
+  class ConfigManager
+    class << self
+      # @return [Hash<Symbol, Config> register of hold configs.
```
ancorgs, line 25: hold -> held

src/lib/y2users/config_manager.rb
```diff
@@ -0,0 +1,48 @@
...
+      def system(reader: nil, force_read: false)
+        res = configs[:system]
+        return res if res && !force_read
+
+        if !reader
+          require "y2users/linux/reader"
+          reader = Linux::Reader.new
+        end
+
+        require "y2users/config"
```
ancorgs, line 39: Why cannot this require be at the top of the file?

jreidinger: probably no strong reason, just only method that needs to know it. But I am fine with moving it up.

ancorgs:
In general looks good. I just wonder whether we want to keep it with class methods or we want a more "classical" singleton class. I don't have a strong preference.

jreidinger:
@ancorgs my only reason to not use Singleton mixin is to avoid calling it with instance

src/lib/y2users/config_manager.rb
```diff
@@ -0,0 +1,48 @@
...
+        res = Config.new
+        reader.read_to(res)
+        configs[:system] = res
+
+        res
+      end
```
joseivanlopez, line 45: In general, I would try to hide the hash under the hood. I would like to have something like:

~~~
manager = ConfigManager.new
config1 = Config.new
config2 = Config.new
manager.register(config1, as: "staging1")
manager.register(config2, as: "staging2")
staging = manager.retrieve("staging2")
manager.configs #=> [config1, config2]
manager.config_names #=> ["staging1", "staging2"]
manager.unregister("staging1")
manager.configs #=> [config2]
~~~

jreidinger: what is benefit of hiding that hash? better DSL? does it pays off? if it is hash it is simple and everyone is familiar with it

jreidinger:
@joseivanlopez @ancorgs this PR should be now ready as we agreed + I add also unit tests ( beside that system method, which I think is useful, but as we do not use it in production yet, I do not consider it final )

jreidinger:
hmm, test failures looks not related to this change

src/lib/y2users/config_manager.rb
```diff
@@ -0,0 +1,84 @@
...
+      @register = {}
+    end
+
+    # Register config as given id
+    # @param [Config] config
+    # @param [Symbol] as id of config
+    # @note if given id is already registered, it is overwritten
+    # rubocop:disable Naming/UncommunicativeMethodParamName
+    def register(config, as:)
+      @register[as] = config
```
ancorgs, line 39: Just to prevent errors, I would ensure `as` is a symbol here. Either by doing `register[as.to_sym] = ...` or by raising an `ArgumentError` if it's not. Otherwise I fear the situation we have in other parts in which we happily mix strings and symbols as keys.

ancorgs: And I would do something similar in `#unregister` and `#config`

src/lib/y2users/config_manager.rb
```diff
@@ -0,0 +1,84 @@
...
+    # returns config for given id
+    # @param [Symbol] id of config
+    # @return [Config,nil] config registered for given id or nil if not found
+    def config(id)
+      @register[id]
+    end
+
+    # Gets system configuration.
+    # @param [#read_to(config)] reader used to read system configuration.
+    #   If not specified it will decide itself to use.
```
ancorgs, line 64: it will decide itself *which one* to use.

ancorgs:
I would also add this new class to `y2users.rb`, I expect it to be the most common entry point to the `Y2Users` classes for most Ruby components.

src/lib/y2users/config_manager.rb
jreidinger: I would probably use Argument error and discuss if we should allow more in future as in general any object can be used.

src/lib/y2users/config_manager.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    def initialize
+      @register = {}
+    end
+
+    # Register config as given id
+    # @param [Config] config
+    # @param [Symbol] as id of config
+    # @note if given id is already registered, it is overwritten
+    # rubocop:disable Naming/UncommunicativeMethodParamName
```
joseivanlopez, line 37: NP: What about configuring rubocop to accept two characters length parameters? By default, `MinNameLength` is 3, see https://www.rubydoc.info/gems/rubocop/0.53.0/RuboCop/Cop/Naming/UncommunicativeMethodParamName.

test/lib/y2users/config_manager_test.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  describe "#register" do
+    it "stores given config" do
+      config1 = Y2Users::Config.new
+      subject.register(config1, as: :test)
+      expect(subject.config(:test)).to be config1
+    end
+
+    it "overwrite previously stored config with same id" do
```
joseivanlopez, line 41: Typo: overwrite**s**

src/lib/y2users/config_manager.rb
jreidinger: I am not sure. Usually I prefer to not deviate unless there are good reason. It is also possible to add "as" as recognized names, I am just not sure if it is generally good name beside this use case.

joseivanlopez (approved):
LGTM

ancorgs (approved):
LGTM

jreidinger:
@ancorgs @joseivanlopez thanks

src/lib/y2users/config_manager.rb
```diff
@@ -0,0 +1,86 @@
...
+    def register(config, as:)
+      raise ArgumentError, "#{as.inspect} is not Symbol" unless as.is_a?(:Symbol)
```
dgdavid, line 39: [`#is_a?`](https://ruby-doc.org/core-3.0.1/Object.html#method-i-is_a-3F) actually expects a class or module. So, it should be

```diff
- raise ArgumentError, "#{as.inspect} is not Symbol" unless as.is_a?(:Symbol)
+ raise ArgumentError, "#{as.inspect} is not Symbol" unless as.is_a?(Symbol)
```
I fixed it in https://github.com/yast/yast-users/pull/259

PR #261, opened 2021-04-30
Adaptation of Dialogs::InstUserFirst to use Y2Users

ancorgs:
Part of https://trello.com/c/nFDiHYCU/2415-validation-of-the-new-user-with-a-better-api

Included:

- Fixed `Y2Users::UsersSimple::Reader`
- Adapted the InstUserFirst dialog to:
  - When the dialog opens:
    - Load the information from `UserSimple` into `Y2User` objects (using the fixed reader)
    - Preload the form with the information from those objects
  - When the user click "next"
    - Update the `Y2User` objects with the information entered by the user
    - Check objects with the new `Y2User` validators, reporting `Y2Issues::Issues` to the user
    - Write the information in the `Y2Users` objects back to `UserSimple` if the objects are valid
- Small reorganization of the code that generates the help texts
  - Better sharing between several dialogs with no direct dependency on `UsersSimple`
  - Allowed to re-enable Rucobop rules for "Metrics/MethodLength" and "Metrics/AbcSize"

Excluded:

- [Rewriting the validations in Ruby](https://trello.com/c/yDXqaGHa/4690-ruby-based-validators), since the current validators still rely on `Yast::UsersSimple` under the hood
- Cleanup of the form to completely stop using `UsersSimple` data structures. For that, we would need to also [migrate to `Y2Users` the functionality of importing users](https://trello.com/c/K0OVGo6g/4674-importing-users-during-installation-also-based-on-new-classes) from previous installations.

Testing:

- Tested manually executing the dialog stand-alone
- Tested in a real installation system (going back, etc.)

### DIFF
--- a/src/lib/users/dialogs/inst_user_first.rb
+++ b/src/lib/users/dialogs/inst_user_first.rb
@@ -20,8 +20,9 @@
 require "yast"
 require "ui/installation_dialog"
 require "users/dialogs/users_to_import"
-require "users/ca_password_validator"
-require "users/local_password"
+require "y2users"
+require "y2users/users_simple"
+require "y2users/help_texts"
 require "users/users_database"
 require "tmpdir"
 
@@ -74,11 +75,9 @@ module Yast
         end
       end
 
-      @users = UsersSimple.GetUsers
       init_action
-      @usernames_to_import = @users.map { |u| u["uid"] || "" } if action == :import
-      @user = @users.first if action == :new_user
-      @user ||= {}
+      # Y2Users: commented until we implement importing
+      # @usernames_to_import = @users.map { |u| u["uid"] || "" } if action == :import
       init_user_attributes
     end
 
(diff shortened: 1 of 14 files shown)


src/lib/y2users/users_simple/reader.rb
```diff
@@ -34,18 +34,25 @@ def read_to(config)
         users = Yast::UsersSimple.GetUsers
         # TODO: only created users, not imported ones for now
         users.each do |user|
-          user = User.new(config, user["uid"], gecos: [user["cn"]])
+          user = User.new(user["uid"])
```
dgdavid, line 37: We're overwriting _user_ here, which is going to raise am `undefined method "[]" for #<Y2Users::User....`

src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -244,40 +175,40 @@ def enable_local_users?
...
-    #
-    # Requires @user to be set in advance
     def init_user_attributes
-      @user_type = @user.fetch("type", "local")
-      @username = @user.fetch("uid", "")
-      @full_name = @user.fetch("cn", "")
-      @password = @user["userPassword"]
+      @username = user&.name || ""
+      @full_name = user&.full_name || ""
+      @password = user&.password&.plain || ""
```
dgdavid, line 193:
```suggestion
      @password = user&.password&.value&.content || ""
```

src/lib/y2users/password_validator.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    # @return [Y2Issues::List]
+    def issues
+      list = Y2Issues::List.new
+
+      return list unless password&.value&.plain?
+
+      err = Yast::UsersSimple.CheckPassword(content, "local")
+      list << Y2Issues::Issue.new(error, location: LOCATION, severity: :fatal) unless err.empty?
```
dgdavid, line 47:
```suggestion
      list << Y2Issues::Issue.new(err, location: LOCATION, severity: :fatal) unless err.empty?
```

dgdavid:
@ancorgs,

While testing changes in the installer, I noticed that validator loads _cracklib-dict-full.rpm_ even when password is empty. See following video:

https://user-images.githubusercontent.com/1691872/116973321-3fd5d080-acb4-11eb-9932-33651b4898c7.mp4

I have to check if it was the behavior before introduced changes, but IMO it worth changing it since we know in advance that the required form password field is empty and there is nothing to validate.

BTW, maybe we could introduce a _UserFormValidator_ or similar for validating the UI before running the _backend_ validators.

ancorgs:
> I have to check if it was the behavior before introduced changes

No, it's not. The problem is introduced in this pull request.

ancorgs:
> No, it's not. The problem is introduced in this pull request.

It's should be fixed now (in a separate commit for clarity).

dgdavid (approved):
LGTM, but please **do not merge it yet**. I'm going to polish the Makefile a little bit.

dgdavid:
> I'm going to polish the Makefile a little bit.
Done

dgdavid:
@ancorgs 

Now the PR is ready to be merged.

PR #262, opened 2021-04-30
Hot fixes

joseivanlopez:
## Problem

Previous PR implements the new API, but it was not tested at all. There were some more than evident bugs.

* https://github.com/yast/yast-users/pull/258

## Solution

Hot-fixes for detected bugs after manual testing. Unit tests will come in a separate PR.


ancorgs (approved):
LGTM

PR #263, opened 2021-04-30
Unit tests

joseivanlopez:
## Problem

API of `Y2Users` classes was re-designed, but unit tests were still missing.

## Solution

Add unit tests and fix some issues.

### DIFF
--- a/test/lib/y2users/user_test.rb
+++ b/test/lib/y2users/user_test.rb
@@ -1,0 +1,451 @@
+#!/usr/bin/env rspec
+
+# Copyright (c) [2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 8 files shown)


test/lib/y2users/user_test.rb
```diff
@@ -0,0 +1,451 @@
...
+
+        it "returns nil" do
+          expect(subject.expire_date).to be_nil
+        end
+      end
+
+      context "and the password has expiration" do
+        let(:expiration) { today }
+
+        let(:today) { Date.today }
```
dgdavid, line 204: NP: this can be done in the same line

dgdavid (approved):
Just a tiny NP which we can live :)

LGTM, thanks!

test/lib/y2users/user_test.rb
joseivanlopez: Only to not repeat `Date.today` below in the check.

PR #264, opened 2021-05-04
Add Local reader and separate parsers

jreidinger:
goal is to have Local reader that can read from target directory ( used when importing users ) and it can be also used for getent reader to distinguish between local and remote users.

### DIFF
--- a/src/lib/y2users/linux/local_reader.rb
+++ b/src/lib/y2users/linux/local_reader.rb
@@ -32,7 +32,7 @@ module Y2Users
       end
 
       def read_to(config)
-        config.attach(read_users + read.groups)
+        config.attach(read_users + read_groups)
         # read passwords after user, as user has to exist in advance
         read_passwords(config)
       end
--- a/src/lib/y2users/linux/local_reader.rb
+++ b/src/lib/y2users/linux/local_reader.rb
@@ -17,17 +17,19 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "yast2/execute"
-
 require "y2users/parsers/group"
 require "y2users/parsers/passwd"
 require "y2users/parsers/shadow"
 
 module Y2Users
   module Linux
-    # Reads users configuration from the system using getent utility.
-    class Reader
+    # Reads local users configuration from the system using /etc files.
+    class LocalReader
       include Yast::Logger
+
+      def initialize(source_dir = "/")
+        @source_dir = source_dir
+      end
 
       def read_to(config)
         config.attach(read_users + read.groups)
@@ -37,26 +39,28 @@ module Y2Users
 
     private
 
+      attr_reader :source_dir
+
       def read_users
-        getent = Yast::Execute.on_target!("/usr/bin/getent", "passwd", stdout: :capture)
+        content = File.read(File.join(source_dir,"/etc/passwd"))
         parser = Parsers::Passwd.new
 
-        parser.parse(getent)
+        parser.parse(content)
       end
 
       def read_groups(config)
-        getent = Yast::Execute.on_target!("/usr/bin/getent", "group", stdout: :capture)
+        content = File.read(File.join(source_dir,"/etc/group"))
         parser = Parsers::Group.new
 
-        parser.parse(getent)
+        parser.parse(content)
       end
 
 
       def read_passwords(config)
-        getent = Yast::Execute.on_target!("/usr/bin/getent", "shadow", stdout: :capture)
+        content = File.read(File.join(source_dir,"/etc/shadow"))
         parser = Parsers::Shadow.new
 
-        passwords = parser.parse(getent)
+        passwords = parser.parse(content)
         passwords.each_pair do |name, password|
           user = config.users.find { |u| u.name == name }
           if !user
--- a/src/lib/y2users/linux/local_reader.rb
+++ b/src/lib/y2users/linux/local_reader.rb
@@ -42,22 +42,21 @@ module Y2Users
       attr_reader :source_dir
 
       def read_users
-        content = File.read(File.join(source_dir,"/etc/passwd"))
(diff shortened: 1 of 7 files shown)


src/lib/y2users/parsers/group.rb
```diff
@@ -0,0 +1,48 @@
...
+        "users"  => 3
+      }.freeze
+
+      # Parses content and returns groups defined there.
+      # @param content [String]
+      # @return [Array<Group>]
+      def parse(content)
+        content.lines.map do |line|
+          values = line.chomp.split(":")
+          group = Group.new(values[GROUP_MAPPING["name"]])
```
dgdavid, line 40:
```suggestion
          group = Y2Users::Group.new(values[GROUP_MAPPING["name"]])
```

src/lib/y2users/parsers/group.rb
```diff
@@ -0,0 +1,48 @@
...
+      GROUP_MAPPING = {
+        "name"   => 0,
+        "passwd" => 1,
+        "gid"    => 2,
+        "users"  => 3
+      }.freeze
+
+      # Parses content and returns groups defined there.
+      # @param content [String]
+      # @return [Array<Group>]
```
dgdavid, line 36:
```suggestion
      # @return [Array<Y2Users::Group>]
```

src/lib/y2users/parsers/passwd.rb
```diff
@@ -0,0 +1,55 @@
...
+        "uid"    => 2,
+        "gid"    => 3,
+        "gecos"  => 4,
+        "home"   => 5,
+        "shell"  => 6
+      }.freeze
+
+      # Parses content and returns users defined there without password value.
+      # @param content [String]
+      # @return [Array<User>]
```
dgdavid, line 39:
```suggestion
      # @return [Array<Y2Users::User>]
```

src/lib/y2users/parsers/shadow.rb
```diff
@@ -0,0 +1,87 @@
...
+        "minimum_age"        => 3,
+        "maximum_age"        => 4,
+        "warning_period"     => 5,
+        "inactivity_period"  => 6,
+        "account_expiration" => 7
+      }.freeze
+
+      # Parses content and returns mapping of username and password there.
+      # @param content [String]
+      # @return [Hash<String, Password>]
```
dgdavid, line 42:
```suggestion
      # @return [Hash<String, Y2Users::Password>]
```

src/lib/y2users/parsers/group.rb
jreidinger: yeah, I get it. Namespace collision

src/lib/y2users/parsers/passwd.rb
jreidinger: Is it really needed? yardoc cannot find it?

dgdavid: I added them just for consistent with the previous one, which had the collision problem.

src/lib/y2users/linux/local_reader.rb
```diff
@@ -0,0 +1,75 @@
...
+    # Reads local users configuration from the system using /etc files.
+    class LocalReader
+      include Yast::Logger
+
+      def initialize(source_dir = "/")
+        @source_dir = source_dir
+      end
+
+      def read_to(config)
+        config.attach(read_users + read.groups)
```
ancorgs, line 35: read_groups instead of read.groups

jreidinger: tests will find it :)

src/lib/y2users/linux/local_reader.rb
```diff
@@ -0,0 +1,75 @@
...
+        read_passwords(config)
+      end
+
+    private
+
+      attr_reader :source_dir
+
+      def read_users
+        content = File.read(File.join(source_dir, "/etc/passwd"))
+        parser = Parsers::Passwd.new
```
joseivanlopez, line 46: NP: I would pass either `content` or the file path to the constructor. I would prefer to associate each parser object to a specific file, so `parser#parse` always returns the same result. Moreover, it will make easier to have local attributes storing e.g. partial results or whatever for the content being parsed.

src/lib/y2users/linux/local_reader.rb
```diff
@@ -0,0 +1,75 @@
...
+# find current contact information at www.suse.com.
+
+require "y2users/parsers/group"
+require "y2users/parsers/passwd"
+require "y2users/parsers/shadow"
+
+module Y2Users
+  module Linux
+    # Reads local users configuration from the system using /etc files.
+    class LocalReader
```
joseivanlopez, line 27: `Linux::LocalReader` and `Linux::Reader` are quite similar. The only one difference is how they get the sources (i.e, reading a file or calling to *getent*). I would suggest to unify them (e.g., using a constructor parameter to indicate the source?) or to create a base class or something that avoids the repetition in the code.

joseivanlopez (commented):
I like the idea of parsers. I have added a couple of suggestions.

src/lib/y2users/linux/local_reader.rb
jreidinger: Well, it start differs as Linux::Reader will call LocalReader to get source of users/groups/etc. I just do not write it yet

src/lib/y2users/linux/local_reader.rb
jreidinger: well, I try to have parser just single responsibility. Reading file and potentially handle IO exceptions for it I would like to have one level up, as e.g. getent does not have it, but File.read can have.
About having parser associated with content I also consider it, but do not see strong reasons for it and if parser is stored somewhere it can mean storing also content in memory which can be quite big on system with multiple users. That is why in the end I prefer just passing it one time.

ancorgs (approved):
LGTM

joseivanlopez (approved):
LGTM

PR #265, opened 2021-05-06
WIP: import users based on Y2Users

ancorgs:


src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -283,9 +263,18 @@ def root_user
       users_config.users.find(&:root?)
     end
 
-    # TODO: let's simply return false for the time being
-    def imported?(_user)
-      false
+    def user_imported?
+      user&.source == :local
```
jreidinger, line 267: I am not sure if this will stay true. As common installation also creates local users.

src/lib/users/users_database.rb
```diff
@@ -53,33 +59,51 @@ def push(database)
...
+      dir = File.join(root_dir, "etc")
       passwd = File.join(dir, "passwd")
       shadow = File.join(dir, "shadow")
       return unless File.exist?(passwd) && File.exist?(shadow)
 
-      self.passwd = IO.read(passwd)
-      self.shadow = IO.read(shadow)
+      @config = Y2Users::Config.new
+      Y2Users::Linux::LocalReader.new(root_dir).read_to(config)
+      config.users.each { |u| u.source = :local }
```
jreidinger, line 96: this in fact should be set in LocalReader. I forgot to have it there, but it makes sense and especially if we want that merge to Reader result.

src/lib/y2users/user.rb
```diff
@@ -213,8 +213,9 @@ def password_issues
     # @return [Boolean]
     def system_uid?
       return false unless uid && sys_uid_max
+      return false if uid == ""
 
-      uid.between?(sys_uid_min || 1, sys_uid_max)
+      uid.to_i.between?(sys_uid_min || 1, sys_uid_max)
```
jreidinger, line 218: I already face this confusion few times, maybe it is time to reconsider having uid as string  and not Integer

src/lib/users/dialogs/inst_user_first.rb
ancorgs: Yes, this is a temporary solution and I wanted to discuss (after some manual testing) what was the best option. So far I fear we are ignoring the `#source` attribute and we should maybe even rethink it.

dgdavid (approved):
LGTM

PR #266, opened 2021-05-06
Adapt Users::PasswordWidget to use Y2Users

dgdavid:
Related to 

* Trello card (internal link): https://trello.com/c/dsDBpo7B/
* Previous PR: https://github.com/yast/yast-users/pull/261

It adapts the `Users::PasswordWidget`, used by _inst\_root\_first_ dialog and widget, to rely on Y2Users objects instead of keep using Yast::UsersSimple directly. 

It has been tested manually, apart to adapt the existent unit tests too.

### DIFF
--- a/src/lib/users/widgets.rb
+++ b/src/lib/users/widgets.rb
@@ -94,9 +94,7 @@ module Users
       Yast::UI.ChangeWidget(Id(:pw2), :Value, current_password)
     end
 
-    # rubocop:disable Metrics/AbcSize
     # rubocop:disable Metrics/CyclomaticComplexity
-    # rubocop:disable Metrics/PerceivedComplexity
     def validate
       password1 = Yast::UI.QueryWidget(Id(:pw1), :Value)
       password2 = Yast::UI.QueryWidget(Id(:pw2), :Value)
@@ -126,9 +124,7 @@ module Users
 
       true
     end
-    # rubocop:enable Metrics/AbcSize
     # rubocop:enable Metrics/CyclomaticComplexity
-    # rubocop:enable Metrics/PerceivedComplexity
 
     def store
       return if allow_empty? && empty?
@@ -137,8 +133,7 @@ module Users
       root_user.password = Y2Users::Password.create_plain(password1)
     end
 
-    # rubocop:disable Metrics/MethodLength
-    def help
+    def help # rubocop:disable Metrics/MethodLength
       # help text ( explain what the user "root" is and does ) 1
       helptext = _(
         "<p>\n" \
@@ -180,7 +175,6 @@ module Users
 
       helptext << ca_password_text
     end
-    # rubocop:enable Metrics/MethodLength
 
   private
 
--- a/src/lib/users/widgets.rb
+++ b/src/lib/users/widgets.rb
@@ -20,9 +20,8 @@
 require "yast"
 require "cwm/widget"
 
-require "y2users"
-require "y2users/users_simple"
 require "y2users/help_texts"
+require "y2users/inst_users_dialog_helper"
 require "users/local_password"
 
 Yast.import "Popup"
@@ -33,6 +32,8 @@ module Users
   # The widget contains 2 password input fields
   # to type and retype the password
   class PasswordWidget < CWM::CustomWidget
+    include Y2Users::InstUsersDialogHelper
+
     class << self
       attr_accessor :approved_pwd
     end
@@ -176,8 +177,6 @@ module Users
       helptext << ca_password_text
     end
 
-  private
-
     # Determines whether the widget is empty or not
     #
     # @return [Boolean]
@@ -186,6 +185,8 @@ module Users
       pw2 = Yast::UI.QueryWidget(Id(:pw2), :Value)
       pw1.to_s.empty? && pw2.to_s.empty?
     end
+
+  private
 
     # Determines whether is allowed to do not fill the password
     #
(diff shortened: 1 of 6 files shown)


src/lib/users/widgets.rb
```diff
@@ -86,16 +89,13 @@ def init
...
-      return if !current_password || current_password.empty?
+      return if current_password.empty?
 
       Yast::UI.ChangeWidget(Id(:pw1), :Value, current_password)
       Yast::UI.ChangeWidget(Id(:pw2), :Value, current_password)
     end
 
-    # rubocop:disable Metrics/AbcSize
     # rubocop:disable Metrics/CyclomaticComplexity
-    # rubocop:disable Metrics/PerceivedComplexity
```
jreidinger, line 98: :+1:

src/lib/users/widgets.rb
```diff
@@ -210,5 +196,22 @@ def empty?
     def allow_empty?
       @allow_empty
     end
+
+    # Current password value for root_user
+    #
+    # @return [String]
+    def current_password
+      pwd = root_user.password&.value&.content
```
jreidinger, line 204: I am just not sure if it will be used also during firstboot? If so, then there is chance that root password is preset to some value and in such case can it be encrypted? Can such case happen?

src/lib/y2users/inst_users_dialog_helper.rb
```diff
@@ -0,0 +1,129 @@
...
+require "yast"
+require "y2users"
+require "y2users/users_simple"
+
+module Y2Users
+  # Mixin for user and password validations based on Y2Users for isntallation dialogs
+  module InstUsersDialogHelper
+    Yast.import "UI"
+    Yast.import "Popup"
+    Yast.import "Report"
```
jreidinger, line 29: I would move that imports to same level as require.

src/lib/y2users/inst_users_dialog_helper.rb
```diff
@@ -0,0 +1,129 @@
...
+
+    # Needed to be able to call textdomain below
+    extend Yast::I18n
+
+    def self.included(_mod)
+      # Needed to prevent the automatic checks (rake check:pot) from complaining
```
jreidinger, line 35: well, it is not just about automatic bots. It has real reason. gettext we use to extract localized messages for translators use textdomain call in given file to associate textdomain with text. That allows us to have more textdomain in single repo ( like Yast2 ) So if it is not there, pot file will miss messages from this file.

src/lib/y2users/inst_users_dialog_helper.rb
```diff
@@ -0,0 +1,129 @@
...
+      @users_config = Y2Users::Config.new
+      Y2Users::UsersSimple::Reader.new.read_to(@users_config)
+      @users_config
+    end
+
+    # All users to be created
+    #
+    # @return [Array<Y2Users::User>]
+    def users
+      users_config.users.reject(&:root?)
```
jreidinger, line 56: UsersSimple does not have root itself, so can it happen here?

jreidinger: ah, I forgot abouut password extraction. Like if I say nothing.

test/lib/y2users/password_validator_test.rb
```diff
@@ -0,0 +1,113 @@
...
+
+  before do
+    allow(user).to receive(:password).and_return(password)
+  end
+
+  describe "#issues" do
+    context "when validating an encrypted password" do
+      let(:password) { Y2Users::Password.create_encrypted("") }
+
+      it "return an empty issues list" do
```
jreidinger, line 44: returns

test/widgets_test.rb
```diff
@@ -131,15 +159,16 @@ def stub_widget_value(id, value)
...
+      expect(Y2Users::Password).to_not receive(:create_plain)
       subject.store
     end
 
     it "stores the value if not empty" do
       stub_widget_value(:pw1, "new cool password")
       stub_widget_value(:pw2, "new cool password")
 
-      expect(Yast::UsersSimple).to receive(:SetRootPassword).with("new cool password")
+      expect(Y2Users::Password).to receive(:create_plain).with("new cool password")
```
jreidinger, line 170: It feels to me like a bit of overmocking. what about
```
expect(root_user).to receive(:password=) do |password|
  expect(password.plain?).to eq true
  expect(password.content).to eq "new cool password"
end
```

test/widgets_test.rb
```diff
@@ -131,15 +159,16 @@ def stub_widget_value(id, value)
       stub_widget_value(:pw1, "")
       stub_widget_value(:pw2, "")
 
-      expect(Yast::UsersSimple).to_not receive(:SetRootPassword)
+      expect(Y2Users::Password).to_not receive(:create_plain)
```
jreidinger, line 162: I would do that expect on user object and not that password is not created. so
```
expect(root_user).to_not receive(:password=) 
```

test/widgets_test.rb
```diff
@@ -112,14 +131,23 @@ def stub_widget_value(id, value)
   it "stores its value" do
     stub_widget_value(:pw1, "new cool password")
 
-    expect(Yast::UsersSimple).to receive(:SetRootPassword).with("new cool password")
+    expect(Y2Users::Password).to receive(:create_plain).with("new cool password")
+    expect(root_user).to receive(:password=).with(password)
```
jreidinger, line 135: same as below

test/widgets_test.rb
```diff
@@ -8,6 +8,16 @@ def stub_widget_value(id, value)
...
 
 describe Users::PasswordWidget do
+  let(:root_user) { Y2Users::User.new("root") }
+  let(:password) { Y2Users::Password.create_plain(pwd) }
+  let(:pwd) { "s3cr3T" }
+
+  before do
+    allow(root_user).to receive(:root?).and_return(true)
+    allow(subject).to receive(:root_user).and_return(root_user)
+    allow(Y2Users::Password).to receive(:create_plain).and_return(password)
```
jreidinger, line 18: ugh, this looks like overmock to me. You basically kill any creation of password and just always create pwd content? this is very strange for me

src/lib/y2users/inst_users_dialog_helper.rb
dgdavid: Oh, thanks for the explanation. I'm going to update the comment, which I stolen from https://github.com/yast/yast-users/blob/54a66f550e9ccf0d002e21410238a222166a096d/src/lib/y2users/help_texts.rb#L31-L37

src/lib/users/widgets.rb
dgdavid: Good question. I don't think so. It not make sense to use _inst\_root\_first_ dialog during firstboot now since It is current purpose is to give the system installed and pre-configured to the final user, not to an administrator. Isn't it? 

It would make sense If firstboot were used to perform the installation, splitting current process in two parts as we already has discussed: to install the very minimum in a first step and then continue with firstboot to fully install and configure the system.

Anyway, I propose to check this in the future, once we have the y2users branch in a better shape.

jreidinger (approved):

PR #267, opened 2021-05-06
Adapt users_finish client

joseivanlopez:


src/lib/y2users/config.rb
```diff
@@ -92,6 +92,16 @@ def clone
       config
     end
 
+    # Merges elements (users and groups) with elements from another config
+    #
+    # @note The caller config is modified.
+    #
+    # @return [Config]
+    def merge(other)
```
jreidinger, line 100: maybe as convention I would call it `merge!` so it is clear that it is modified and consistent with Hash.

Another topic is that merged. I would do it as
```
    def merge(other, merger: Mergers::Replace.new)
       merger.merge(self, other)
    end
```
this way you can pass merger with different strategy.

src/lib/y2users/config.rb
```diff
@@ -199,5 +209,92 @@ def self.next
...
+
+        copy_group_values(group, lhs_group)
+
+        lhs.detach(lhs_group)
+        lhs.attach(group)
+      end
+
+      def copy_user_values(user, lhs_user)
+        nil
+      end
```
jreidinger, line 297: Well, for default replacing merger this is overkill. For autoyast one I prefer to have own object with own methods that needs a bit of complexity, but here it is unnecessary.

I expect that something like 
```
def merge(lhs, rhs)
  rhs.users.each_with_object(lhs.users) { |ru, lusers| merge_into(lhs, lusers, ru) }
  rhs.groups.each_with_object(lhs.groups) { |ru, lgroups| merge_into(lhs, lgroups, ru) }
end

def merge_into(config, old_elements, replacing_element)
  new_element = replacing_element.clone
  old_element = old_elements.find { |e| e.name == new_element.name }
  if old_element
    new_element.assign_internal_id(old_element.id) # do we really want this? Does same name really mean edit and not recreate?
     config.detach(old_element)
  end
  config.attach(new_element)
end
  
    
end
```

src/lib/y2users/config.rb
```diff
@@ -199,5 +209,92 @@ def self.next
...
+      def merge_new_group(rhs_group)
+        group = rhs_group.clone
+
+        lhs.attach(group)
+      end
+
+      def merge_existing_group(lhs_group, rhs_group)
+        group = rhs_group.clone
+
+        group.assign_internal_id(lhs_group.id)
```
jreidinger, line 287: I am not sure if same name always means just edit. What is reason for this?

src/lib/y2users/config.rb
```diff
@@ -92,6 +93,16 @@ def clone
       config
     end
 
+    # Merges elements (users and groups) with elements from another config
+    #
+    # @note The caller config is modified.
+    #
+    # @return [Config]
```
jreidinger, line 100: please document parameters

src/lib/y2users/config.rb
```diff
@@ -92,6 +93,16 @@ def clone
...
+    def merge(other, merger: nil)
+      merger ||= ConfigMergers::Replace
+      merger.new(self, other).merge
```
jreidinger, line 103: NP: I still find better to pass arguments to merge instead of constructor, but not critical.

src/lib/y2users/config_mergers.rb
```diff
@@ -0,0 +1,24 @@
...
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Y2Users
+  module ConfigMergers; end
```
jreidinger, line 21: If you define it here, I think it would be nice to add documenatation to module.

src/lib/y2users/config_mergers/base.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    private
+
+      # @return [Config] Left Hand Size config
+      attr_reader :lhs
+
+      # @return [Config] Right Hand Size config
+      attr_reader :rhs
+
+      abstract_method :merge_element
```
jreidinger, line 49: to be honest I do not like this. E.g. Autoyast merger needs specific strategy for users and groups, so then I would need in abstract method to again separate them. Why not simply make merge method abstract?

src/lib/y2users/config_mergers/replace.rb
```diff
@@ -0,0 +1,41 @@
...
+      def merge_element(config, element)
+        current_element = find_element(config, element)
+
+        config.detach(current_element) if current_element
+
+        config.attach(element.clone)
+      end
+
+      def find_element(config, element)
+        elements = element.is_a?(User) ? config.users : config.groups
```
jreidinger, line 35: this is problematic if we add more config attributes in future. If you want to do it this way, I would use case statement with log/exception in else branch.

src/lib/y2users/config.rb
joseivanlopez: IMHO, having the objects as internal attributes makes the implementation of the mergers more clear and easy to understand.

src/lib/y2users/config_mergers/base.rb
joseivanlopez: Well, I think that `#merge` itself is always the same: to iterate all elements and to merge each element into `lhs`. So, for me, it is something shared by all the mergers. Only the logic to merge an element into a config is what changes among the mergers. On other hand, if autoyast requires to distinguish the type of element, I guess it can do it just inside its `#merge_element` method.

src/lib/y2users/config.rb
joseivanlopez: Ok, I think that copying the id should not be the default for `ConfigMergers::Replace`. But, most likely we will need to preserve the id when we are merging users_simple config into the system config. If so, we can add a new `ConfigMergers::UsersSimple`.

joseivanlopez:
Closed in favor of https://github.com/yast/yast-users/pull/268.

PR #268, opened 2021-05-07
Add ConfigMerger

joseivanlopez:
## Problem

In the *users_finish* client, we need to pass two configs to the config writer in order to create the new users. The idea is to read the users from the system, and then to generate a new config from the system one but adding all the user that were stored in *UsersSimple*. Then, the writer would take care of creating the new users by comparing these two configs. Note that the password of root user should be updated if needed.

In order to keep the writer as simple as possible, we need to prepare the target config in a way that the new users can be created and the root password can be updated by the writer. At the same time, the rest of users from system should not be touched at all.

## Solution

Create a `ConfigMerger` class to merge the users and groups from one config into another config. This will be useful in other scenarios too, for example AutoYaST.

### DIFF
--- a/src/lib/y2users/config_merger.rb
+++ b/src/lib/y2users/config_merger.rb
@@ -1,0 +1,105 @@
+# Copyright (c) [2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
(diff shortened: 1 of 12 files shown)


src/lib/y2users/config_merger.rb
```diff
@@ -0,0 +1,105 @@
...
+    # that for users, their uid and gid are not updated according to the {rhs} values.
+    def merge
+      elements = rhs.users + rhs.groups
+
+      elements.each { |e| merge_element(lhs, e) }
+    end
+
+  private
+
+    # Left Hand Size config
```
dgdavid, line 44: Size? Do you mean _Side_? If so, please update all occurrences.

src/lib/y2users/config_merger.rb
```diff
@@ -0,0 +1,105 @@
...
+      else
+        raise "Element #{element} not valid. It must be an User or Group"
+      end
+
+      elements.find { |e| e.name == element.name }
+    end
+
+    # Copies values from one element into another
+    #
+    # @param from [Config] source element
```
dgdavid, line 96: Are we actually expecting a Config here? Or an element?

dgdavid (approved):
LGTM, just minor documentation issues

PR #269, opened 2021-05-10
Improved reader and writer for UserSimple

ancorgs:
Extracted from #265 

This PR includes:

 - Improvements in `UsersSimple::Writer` to properly write both encrypted and plain passwords and also to write all the shadow attributes (expiration date, last password change, etc.)
 - Decent implementation of `UsersSimple::Reader` supporting all attributes (the previous version was very limited).
 - Bonus track: updated Makefile (it didn't include several of the files added in previously merged pull requests).

### DIFF
--- a/src/lib/y2users/users_simple/reader.rb
+++ b/src/lib/y2users/users_simple/reader.rb
@@ -20,6 +20,7 @@
 require "yast"
 require "y2users/user"
 require "y2users/password"
+require "y2users/parsers/shadow"
 
 Yast.import "UsersSimple"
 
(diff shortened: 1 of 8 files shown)


src/lib/y2users/users_simple/reader.rb
```diff
@@ -20,13 +20,21 @@
...
 
 Yast.import "UsersSimple"
 
 module Y2Users
   module UsersSimple
     # Class for reading users configuration from old Yast::UsersSimple module
     class Reader
+      # @see #{shadow_string}
+      SORTED_SHADOW_ATTRS = [
+        "uid", "userPassword", "shadowLastChange", "shadowMin", "shadowMax",
```
jreidinger, line 33: I would write here that uid is actually username as it is confusing as shadow does not contain uid.

jreidinger (approved):

PR #270, opened 2021-05-10
Adapt users_finish client

joseivanlopez:
### Problem

Clients *inst_user_first* and *inst_root_first* were already adapted to use the new `Y2Users` classes, but *users_finish* client is still using the old code to create the users into the system.

### Solution

The *users_finish* client was adapted to use `Y2Users` classes.

NOTE: *users_finish* is also used by AutoYaST, but the AutoYaST part will be adapted in a separate PR.

## Testing

* Added unit tests
* Manually tested:
  * Autologin is not working (will be fixed in a separate PR)
  * Root password form has an issue: https://github.com/yast/yast-users/pull/273

### DIFF
--- a/src/lib/users/clients/users_finish.rb
+++ b/src/lib/users/clients/users_finish.rb
@@ -1,17 +1,28 @@
-# ------------------------------------------------------------------------------
-# Copyright (c) 2016 SUSE LLC
+# Copyright (c) [2016-2021] SUSE LLC
 #
-# This program is free software; you can redistribute it and/or modify it under
-# the terms of version 2 of the GNU General Public License as published by the
-# Free Software Foundation.
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
 #
 # This program is distributed in the hope that it will be useful, but WITHOUT
-# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
-# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
 #
-# ------------------------------------------------------------------------------
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
 
+require "yast"
 require "installation/finish_client"
+require "y2users/config"
+require "y2users/config_manager"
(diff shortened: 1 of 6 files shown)


test/users_finish_test.rb
```diff
@@ -71,29 +74,87 @@
...
+          it "reports all errors to the user" do
+            expect(Yast::Report).to receive(:Error).with("error 1\n\nerror 2")
+
+            subject.run
+          end
         end
 
-        context "when a new user is added" do
-          let(:user_added) { true }
+        context "when an user is configured for auto-login" do
```
dgdavid, line 138: NP: I think it is [_**a** user_](https://english.stackexchange.com/questions/105116/is-it-a-user-or-an-user). And I think now that probably I wrote _an user_ in the past and/or did not catch them in previous reviews.

src/lib/users/clients/users_finish.rb
```diff
@@ -82,12 +93,86 @@ def write_autoinst
       Progress.set(@progress_orig)
     end
 
-    # Write root password a new users during regular installation
+    # Writes users during the installation
+    #
+    # The linux writer will proceed all the differences between the system and the target configs.
```
ancorgs, line 98: proceed -> process

src/lib/users/clients/users_finish.rb
```diff
@@ -82,12 +93,86 @@ def write_autoinst
...
+    def system_config
+      config_manager.system
+    end
+
+    # Target config, which extends the system config with the new users that should be created
+    # during the installation.
+    #
+    # @return [Y2Users::Config]
+    def target_config
+      target = config_manager.config(:target)
```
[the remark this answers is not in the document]
joseivanlopez: No real benefit, but as we need a ConfigMaganer in order to retrieve system ... But we can avoid to store *target* there.

src/lib/users/clients/users_finish.rb
```diff
@@ -82,12 +93,86 @@ def write_autoinst
...
+
+      Autologin.user = UsersSimple.GetAutologinUser
+      Autologin.Use(true)
+    end
+
+    # System config, which contains all the current users on the system
+    #
+    # @return [Y2Users::Config]
+    def system_config
+      config_manager.system
```
ancorgs, line 133: I wonder if we should use `force_read` here.

src/lib/users/clients/users_finish.rb
```diff
@@ -82,12 +93,86 @@ def write_autoinst
...
+      Y2Users::ConfigManager.instance
+    end
+
+    # Reports issues
+    #
+    # @param issues [Array<Y2Issues::Issue>]
+    def report_issues(issues)
+      message = issues.map(&:message).join("\n\n")
+
+      Report.Error(message)
```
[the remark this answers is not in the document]
joseivanlopez: Created https://trello.com/c/08t84wmL/4716-a-better-way-for-reporting-issues-during-installation.

ancorgs (approved):
LGTM

PR #272, opened 2021-05-11
Add basic AutoYaST support

jreidinger:
This PR implements support to import and write users during autoinstallation. Only the 1st stage is supported.

Trello: https://trello.com/c/uSZUbp5E/2409-5-autoyast-during-installation-write-users-using-y2users

## TODO

* [x] Write tests for `users_finish`.
* [x] Merge `y2users` branch because it contains fundamental changes


src/Makefile.am
```diff
@@ -116,6 +116,16 @@ ylib_y2users_userssimple_DATA = \
   lib/y2users/users_simple/reader.rb \
   lib/y2users/users_simple/writer.rb
 
+ylib_y2users_autoyastdir = @ylibdir@/y2users/autoyast
+ylib_y2users_autoyast_DATA = \
+  lib/y2users/autoyast/hash_reader.rb
+
+ylib_y2users_parsersdir = @ylibdir@/y2users/parsers
```
joseivanlopez, line 123: Parsers are already added above.

src/lib/users/clients/users_finish.rb
```diff
@@ -76,21 +80,19 @@ def write_autoinst
...
       # 4. Write users
-      Users.SetWriteOnly(true)
-      @progress_orig = Progress.set(false)
-      error = Users.Write
-      log.error(error) unless error.empty?
-      Progress.set(@progress_orig)
+      # TODO: Write login defaults only
+      issues = Y2Users::Linux::Writer.new(merged_config, system_config).write
+      log.error(issues.inspect)
+      report_issues(issues) if issues.any?
```
dgdavid, line 95: NP: only log if there are errors.

```suggestion
      return if issues.empty? 
      
      log.error(issues.inspect)
      report_issues(issues)
```

src/lib/y2users/autoyast/hash_reader.rb
```diff
@@ -0,0 +1,103 @@
...
+      end
+
+      # shadow attrs without password and username which is done manually
+      SORTED_SHADOW_ATTRS = [
+        # Looks like shadow last change is not part of User.Export TODO: verify
+        "shadowLastChange", "min", "max",
+        "warn", "inact", "expire", "flag"
+      ].freeze
+
+      # Creates a {Password} object based on the data structure of an Users user
```
dgdavid, line 76: _of a user_

src/lib/y2users/config_merger.rb
```diff
@@ -97,15 +101,23 @@ def find_element(config, element)
...
-      # For an user, its uid and gid should not be updated in order to keep the current AutoYaST
-      # behavior. In general, this only affects to the AutoYaST case. In other scenarios like a
-      # regular installation, all users from {rhs} (except root) are new users, so this code is
-      # not executed for them. And, during installation, only the password can be modified for root,
-      # so this actually changes nothing for the root user.
-      to.uid = from.uid
-      to.gid = from.gid
+      case from
+      when User
+        # For an user, its uid and gid should not be updated in order to keep the current AutoYaST
```
dgdavid, line 106:
```suggestion
        # For a user, its uid and gid should not be updated in order to keep the current AutoYaST
```

src/lib/y2users/config_merger.rb
```diff
@@ -97,15 +101,23 @@ def find_element(config, element)
...
+        # behavior. In general, this only affects to the AutoYaST case. In other scenarios like a
+        # regular installation, all users from {rhs} (except root) are new users, so this code is
+        # not executed for them. And, during installation, only the password can be modified for root,
+        # so this actually changes nothing for the root user.
+        to.uid = from.uid
+        to.gid = from.gid
+      when Group
+        # For group it does not modify its group id as it is hard to modify it. Reason is similar as
```
dgdavid, line 114:
```suggestion
        # For a group, it does not modify its group id as it is hard to modify it. Reason is similar as
```

src/lib/y2users/linux/writer.rb
```diff
@@ -101,8 +101,15 @@ def initialize(config, initial_config)
       def write
         issues = Y2Issues::List.new
 
+        # handle groups first as it does not depend on users uid, but it is vice versa
+        new_groups = config.groups.map(&:id) - initial_config.groups.map(&:id)
+        new_groups.map! { |id| config.groups.find { |g| g.id == id } }
+        new_groups.each { |g| add_group(g, issues) }
+
+        # TODO: modify group?
```
dgdavid, line 109: Note: maybe we could include a `GroupFinder` similar to `UserFinder`

src/lib/y2users/linux/writer.rb
```diff
@@ -101,8 +101,15 @@ def initialize(config, initial_config)
...
+
         users_finder.added.each { |u| add_user(u, issues) }
-        users_finder.password_edited.each { |u| change_password(u, issues) }
+        users_finder.modified.each { |nu, ou| modify(nu, ou, issues) }
```
dgdavid, line 112: NP: I'd write `new_user` and `old_user`

src/lib/y2users/linux/writer.rb
```diff
@@ -158,6 +165,68 @@ def add_user(user, issues)
         log.error("Error creating user '#{user.name}' - #{e.message}")
       end
 
+      # Command for creating new users
+      GROUPADD = "/usr/sbin/useradd".freeze
```
dgdavid, line 169: Do you mean [groupadd](https://www.man7.org/linux/man-pages/man8/groupadd.8.html)?

src/lib/y2users/linux/writer.rb
```diff
@@ -158,6 +165,68 @@ def add_user(user, issues)
         log.error("Error creating user '#{user.name}' - #{e.message}")
       end
 
+      # Command for creating new users
```
dgdavid, line 168: Do you mean _for creating new groups_?

src/lib/y2users/linux/writer.rb
```diff
@@ -158,6 +165,68 @@ def add_user(user, issues)
...
+      # @param group [Y2User::Group] the group to be created on the system
+      # @param issues [Y2Issues::List] a collection for adding an issue if something goes wrong
+      def add_group(group, issues)
+        args = [GROUPADD]
+        args << "--gid" << group.gid if group.gid
+        # TODO: system groups?
+        Yast::Execute.on_target!(args)
+      rescue Cheetah::ExecutionFailed => e
+        issues << Y2Issues::Issue.new(
+          format(_("The user '%{username}' could not be created"), username: user.name)
```
dgdavid, line 183:
```suggestion
          format(_("The group '%{groupname}' could not be created"), groupname: group.name)
```

src/lib/y2users/linux/writer.rb
```diff
@@ -158,6 +165,68 @@ def add_user(user, issues)
...
+        )
+        log.error("Error creating user '#{user.name}' - #{e.message}")
```
dgdavid, line 185:
```suggestion
        log.error("Error creating group '#{group.name}' - #{e.message}")
```

src/lib/users/clients/users_finish.rb
```diff
@@ -76,21 +80,19 @@ def write_autoinst
       Users.SetExportAll(false)
       saved = Users.Export
       log.info("Users to import: #{saved}")
-
-      # 2. Read users and settings from the installed system
-      # (bsc#965852, bsc#973639, bsc#974220 and bsc#971804)
-      Users.Read
+      ay_config = Y2Users::Config.new
+      # TODO: support for BTRFS home and also what about login defaults?
+      Y2Users::Autoyast::HashReader.new(saved).read_to(ay_config)
```
joseivanlopez, line 85: NP: passing a config to `#read_to` is something that I don't like too much. IMHO, it would be better if our readers simply generates a new config:

~~~
reader = Y2Users::Linux::Reader.new
config = reader.read
~~~ 

The reader object could even cache the resulting config, and return a clone in the next calls:

~~~
config = reader.read #=> result is internally cached
config2 = reader.read #=> returns a clone of the cached config
config3 = reader.read(force: true) #=> discards the cached result and reads again
~~~

What do you think?

src/lib/y2users/autoyast/hash_reader.rb
```diff
@@ -0,0 +1,103 @@
...
+          res.gid = e_group["gid"]
+          users_name = e_group["userlist"].split(",")
+          res.users_name = users_name
+          res.source = :local
+          res
+        end
+      end
+
+      # shadow attrs without password and username which is done manually
+      SORTED_SHADOW_ATTRS = [
```
joseivanlopez, line 70: I see some duplication here with the `UsersSimple::Reader`. Both have almost the same code for creating the passwords. Maybe we could generalize it.

src/lib/y2users/config_merger.rb
```diff
@@ -97,15 +101,23 @@ def find_element(config, element)
...
+        # for user. Change of gid require manual steps with change of ownership of files which YaST
```
joseivanlopez, line 115: NP: I would reword it a bit: "Changing the group gid requires some manual steps in which the ownership of some files has to be changed. YaST does not want to be responsible of that changes."

src/lib/y2users/linux/writer.rb
```diff
@@ -158,6 +165,68 @@ def add_user(user, issues)
...
+        usermod_modify(new_user, old_user, issues) if usermod_changes
+      end
+
+      def different_groups?(lhu, rhu)
+        lhu.groups(with_primary: false).sort != rhu.groups(with_primary: false).sort
+      end
+
+      USERMOD = "/usr/sbin/usermod".freeze
+      private_constant :USERMOD
+      def usermod_modify(new_user, old_user, issues)
```
dgdavid, line 208: Actually, it does not matter if you run the command with the same previous values, right? So, I'd simplify this method just doing something similar that it was done in useradd_options: include all options with the new_user values, reject those nil values and run the command.


```ruby
opts = {
  "--home"    => new_user.home,
  "--shell"   => new_user.shell,
  "--comment" => new_user.gecos.join(","),
  "--groups"  => new_users.groups(with_primary: false).join(",")
}

opts = opts.compact.flatten
opts << "--move-home" if opts.include?("--home")
opts << new_user.name

Yast::Execute.on_target!(USERMOD, opts)
```

src/lib/y2users/linux/writer.rb
jreidinger: well, we agreed on introduction that collections, so I do not want to do unnecessary work when it is planned to be replaced.

src/lib/y2users/linux/writer.rb
jreidinger: well, my internal rule ( and I think also rubocop has it ) if it is `{}` block so one liner, short names can be used. If it is longer one, I use more descriptive name.

src/lib/y2users/linux/writer.rb
jreidinger: good catch

src/lib/y2users/linux/writer.rb
jreidinger: yep, c&p error

src/lib/y2users/config_merger.rb
```diff
@@ -97,15 +101,23 @@ def find_element(config, element)
...
+        # do not want to be responsible for.
+        to.gid = from.gid
+      else
+        log.info "No specific copy for #{from.inspect}"
```
joseivanlopez, line 119: I don't understand what this log message actually means. I would log something like: "Values can only be copied for a User or Group. Unexpected object type: #{from.inspect}."

src/lib/y2users/user.rb
```diff
@@ -113,15 +113,21 @@ def primary_group
...
-    def groups
+    def groups(with_primary: true)
       return [] unless attached?
 
-      config.groups.select { |g| g.users.include?(self) }
+      res = config.groups.select { |g| g.users.include?(self) }
+
+      primary_group_cache = primary_group
+      res.delete_if { |g| g == primary_group_cache } unless with_primary
+
```
joseivanlopez, line 129: NP: Maybe this is more straightforward?

~~~
groups = config.groups.select { |g| g.users.include?(self) }
groups.reject { |g| g.gid == gid } unless with_primary
~~~

src/lib/y2users/linux/writer.rb
```diff
@@ -158,6 +165,68 @@ def add_user(user, issues)
...
+        args = [USERMOD]
+        if new_user.gecos != old_user.gecos && new_user.gecos
```
dgdavid, line 210: I think the last part of the condition is not needed since `gecos` should be always an array. Emptied or filled array, but not nil.

src/lib/y2users/linux/writer.rb
```diff
@@ -158,6 +165,68 @@ def add_user(user, issues)
...
 
+      # Command for creating new users
+      GROUPADD = "/usr/sbin/useradd".freeze
+      private_constant :USERADD
+
+      # Executes the command for creating the group
+      #
+      # @param group [Y2User::Group] the group to be created on the system
+      # @param issues [Y2Issues::List] a collection for adding an issue if something goes wrong
+      def add_group(group, issues)
```
joseivanlopez, line 176: Thinking loud: Maybe the logic for creating, editing and removing users/groups deserves its own classes, something like `Action::Create`, `Action::Edit`, `Action::Delete`. The writer is growing and adding a lot logic.

src/lib/y2users/linux/writer.rb
dgdavid: Sorry, I didn't know / remind it :innocent:

src/lib/y2users/config_merger.rb
dgdavid: I agree. I was tempted to suggest a similar message than in https://github.com/yast/yast-users/blob/y2users/src/lib/y2users/config_merger.rb#L89, but then I discarded the idea.

src/lib/y2users/linux/writer.rb
jreidinger: well, I am not sure if it matters. E.g. if home is same and option move-home is used if it do something or not. So to be sure I use approach which pass it only when it really is needed.

src/lib/y2users/linux/writer.rb
jreidinger: yeah, probably something to change as it is different if you intentionally keep it empty or use nil for I do not modify.

jreidinger: @ancorgs ^^^

src/Makefile.am
jreidinger: yeah, failure of merge. Will fix it.

src/lib/users/clients/users_finish.rb
jreidinger: It makes sense for me.

src/lib/y2users/linux/writer.rb
dgdavid: I see. Anyway, I did some test before commenting and it looks to me that `usermod` is smart enough to only perform real changes. That said, its man page says nothing to that regard.

src/lib/y2users/autoyast/hash_reader.rb
jreidinger: sadly not. As you can see here is a big inconsistency between attrs names see e.g. min, max..and UsersSimple use ShadowMin and ShadowMax and it is not in nested hash.

src/lib/y2users/config_merger.rb
jreidinger: in such case I would turn it into error if it is not expect object.

src/lib/y2users/user.rb
jreidinger: sadly not. As it is nil if with_primary is false. Try in IRB
```
a = [0,1]
=> [0, 1]
irb(main):003:0> a.reject(&:even?) if false
=> nil
```

src/lib/y2users/linux/writer.rb
jreidinger: if it grow too much, then it is fine for me. But still it is writer specific so it probably needs to be under linux namespace and then `Y2Users::Linux::Action::Create` looks quite long for me.

src/lib/y2users/autoyast/hash_reader.rb
jreidinger: I have to google it why it is correct with sound of "y" https://www.quora.com/Is-it-an-user-or-a-user

src/lib/y2users/linux/writer.rb
dgdavid: >  E.g. if home is same and option move-home is used if it do something or not

Just out of curiosity, I have been taking a look to the [`usermod` source code](https://github.com/shadow-maint/shadow/blob/master/src/usermod.c) and checked that it prevents trying to [move_home](https://github.com/shadow-maint/shadow/blob/ca1d912e6b38c4e065592d4cc61d8365e4210882/src/usermod.c#L2398-L2400) by [overriding the flag](https://github.com/shadow-maint/shadow/blob/3ac8d97825f01055974039dd29eb1fb59c380b23/src/usermod.c#L1423-L1427) when both, given directory is NULL or it is the same.

dgdavid: Also checked that [we're not changing](https://build.opensuse.org/package/show/SUSE:SLE-15-SP3:GA/shadow) its behavior ;)

src/lib/y2users/autoyast/hash_reader.rb
joseivanlopez: Ok, I thought it was only a matter of key names, but if the structure is different ...

src/lib/y2users/user.rb
joseivanlopez: Yes, sorry, what I wanted to propose is to use #gid directly:

~~~
groups = config.groups.select { |g| g.users.include?(self) }
groups.reject! { |g| g.gid == gid } unless with_primary
groups
~~~

src/lib/y2users/linux/writer.rb
joseivanlopez: Inside the writer it would be `Action::Create` only, but it was only a wild idea, nothing to be done now.

src/lib/y2users/user.rb
jreidinger: ah, it makes sense then

src/lib/y2users/linux/writer.rb
jreidinger: I expect we will introduce it when we add also removal of user

src/lib/y2users/autoyast/hash_reader.rb
jreidinger: yes, see https://github.com/yast/yast-users/pull/272/files/f60488738f5aac75e34b295676e45e8405b8ad5f#diff-58cefa9c64c06ad09607ab955a8b672cc20183a654199e76fd3f6ce784c04330R94 it reads that nested hash. I think it is reason why UsersSimple has shadow prefix everywhere to distinguish between shadow and passwd keys.

src/lib/y2users/linux/writer.rb
```diff
@@ -158,6 +165,66 @@ def add_user(user, issues)
...
+          "--shell"   => new_user.shell,
+          "--comment" => new_user.gecos.join(","),
+          "--groups"  => new_users.groups(with_primary: false).join(",")
+        }
+
+        opts = opts.compact.flatten
+        opts << "--move-home" if opts.include?("--home")
+        opts << new_user.name
+
+        Yast::Execute.on_target!(USERMOD, *args)
```
dgdavid, line 220:
```suggestion
        Yast::Execute.on_target!(USERMOD, *opts)
```

dgdavid (changes requested):

src/lib/y2users/linux/writer.rb
```diff
@@ -158,6 +165,66 @@ def add_user(user, issues)
...
+        log.error("Error creating group '#{group.name}' - #{e.message}")
+      end
+
+      USERMOD_ATTRS = [:home, :shell, :gecos].freeze
+
+      def modify(new_user, old_user, issues)
+        change_password(new_user, issues) if new_user.password != old_user.password
+
+        usermod_changes = USERMOD_ATTRS.any? do |attr|
+          !new_user.public_send(attr).nil? &&
```
dgdavid, line 194: NP: Remember that by now `gecos` is never going to be `nil`.

jreidinger: yep, we should probably change that as I write above.

src/lib/y2users/linux/writer.rb
```diff
@@ -175,7 +175,7 @@ def add_user(user, issues)
       # @param issues [Y2Issues::List] a collection for adding an issue if something goes wrong
       def add_group(group, issues)
         args = [GROUPADD]
-        args << "--gid" << group.gid if group.gid
+        args << "--gid" << group.gid if group.gid && valid_gid?(group)
```
jreidinger, line 178: this needs revert as we agreed that it is not writer responsibility.

src/lib/y2users/linux/writer.rb
```diff
@@ -186,6 +186,13 @@ def add_group(group, issues)
         log.error("Error creating group '#{group.name}' - #{e.message}")
       end
 
+      # TODO: do we really want to ignore user configuration? if we pass it, it will fail to create group
+      def valid_gid?(group)
+        same_gids = group.config.groups.select { |g| g.gid == group.gid }
+        # non colliding gid
+        same_gids.size == 1
+      end
```
jreidinger, line 194: and whole this method

src/lib/users/clients/users_finish.rb
imobachgs: Unless we have a use case, I would simply return the new configuration (with no caching at all). I would move the responsibility to hold the config to the caller. But if we decide to cache the result, you can avoid the `force` parameter by just using a new reader.

src/lib/y2users/autoinst/config_merger.rb
```diff
@@ -0,0 +1,98 @@
...
+        end
+
+        config.attach(new_element)
+      end
+
+      # Merges one element into another
+      #
+      # @param from [User, Group] Source element
+      # @param to [User, Group] Target element. This element is modified.
+      def merge_attributes(from, to)
```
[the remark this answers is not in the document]
imobachgs: I do not think so. It is not common to check the class in Ruby. Moreover, where do you put the line them? In which situations would you check it?

src/lib/y2users/autoinst/config_merger.rb
```diff
@@ -0,0 +1,98 @@
...
+        end
+      end
+
+      # Merges two users
+      #
+      # Only non-nil from the `to` user are considered.
+      #
+      # @param [User] Source user
+      # @param [User] Target user. This user is modified.
+      def merge_users(from, to)
```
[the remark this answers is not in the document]
imobachgs: Same than above :smiley:

test/Makefile.am
```diff
@@ -17,6 +17,7 @@ TESTS = \
   lib/y2users/linux/local_reader_test.rb \
   lib/y2users/linux/reader_test.rb \
   lib/y2users/linux/writer_test.rb \
+  lib/y2users/autoinst/reader_test.rb \
```
joseivanlopez, line 20: Please, add *autoinst/config_merger_test.rb* too.

joseivanlopez (approved):
LGTM

PR #273, opened 2021-05-11
Write root password to Yast::UsersSimple

dgdavid:
### Problem

While manually testing https://github.com/yast/yast-users/pull/270, I realize that the root password was not being set as expected because an step missing in https://github.com/yast/yast-users/pull/266: to _write back_ the chosen root password to `Yast::UsersSimple` database.

### Solution

This PR fix it by updating the UserSimple password in the PasswordWidget#store method, as it was before #266, but ideally it should be relocated in a more appropriate place in future iterations.

### Tests

* Updated unit tests
* Tested manually too via DUD and proceeding with installation until reach the installed system

### DIFF
--- a/test/widgets_test.rb
+++ b/test/widgets_test.rb
@@ -143,26 +143,70 @@ describe Users::PasswordWidget do
       expect(root_user).to_not receive(:password_issues)
       expect(subject.validate).to eq(true)
     end
+  end
 
-    it "does not store the value if empty" do
-      stub_widget_value(:pw1, "")
-      stub_widget_value(:pw2, "")
+  describe "#store" do
+    let(:user_simple_writer) { double(Y2Users::UsersSimple::Writer, write: true) }
 
-      expect(root_user).to_not receive(:password=)
-      subject.store
+    before do
+      allow(Y2Users::UsersSimple::Writer).to receive(:new).and_return(user_simple_writer)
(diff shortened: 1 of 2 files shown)


dgdavid:
Now waiting for #274 be merged

test/widgets_test.rb
```diff
@@ -154,26 +142,70 @@ def stub_widget_value(id, value)
...
+      end
     end
 
-    it "stores the value if not empty" do
-      stub_widget_value(:pw1, "new cool password")
-      stub_widget_value(:pw2, "new cool password")
+    context "when the widget is not allowed to be empty" do
+      subject { described_class.new(allow_empty: false) }
+
+      context "and password is empty" do
```
joseivanlopez, line 198: If empty is not allowed, then this should not be possible, right? I mean, it should complain because the password is empty. Or am I missing something?

dgdavid: You're right. I was talking about it with @ancorgs this morning before daily calls. I added them there because I was testing the #store method _in isolation_, but it does not make sense: in the CWM _life cycle_, the store method will be reached only when #validate returns true. And #validate will not allow having empty password if `allow_empty: false`.

So, by now I'm going to simplify them by testing only _real_ behavior.

Anyway, thanks for checking it :)

joseivanlopez (approved):
LGTM

PR #274, opened 2021-05-12
Do not use an attached user for validations

dgdavid:
Stop using a user attached to the config for performing validations because, as written in the commit message 

> [...] could end up with changes that the user
eventually discarded (e.g., the user changes their mind about using the
same password for the root user)

Instead, it creates a detached _placeholder_ user for it.

### DIFF
--- a/src/lib/users/dialogs/inst_user_first.rb
+++ b/src/lib/users/dialogs/inst_user_first.rb
@@ -321,11 +321,12 @@ module Yast
         return false
       end
 
-      user.password = Y2Users::Password.create_plain(pw1)
-      root_user.password = Y2Users::Password.create_plain(pw1) if @use_pw_for_root
+      @password = pw1
       return false unless valid_password?
 
-      @password = pw1
+      user.password = Y2Users::Password.create_plain(pw1)
+      root_user.password = Y2Users::Password.create_plain(pw1) if @use_pw_for_root
+
       true
     end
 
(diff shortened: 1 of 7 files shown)


test/widgets_test.rb
```diff
@@ -13,6 +13,7 @@ def stub_widget_value(id, value)
   before do
     allow(root_user).to receive(:root?).and_return(true)
     allow(subject).to receive(:root_user).and_return(root_user)
+    allow(subject).to receive(:user_to_validate).and_return(root_user)
```
joseivanlopez, line 16: I think that test coverage is decreasing because `#user_to_validate` now has several lines of code and they are all mocked. Maybe we can simply mock `Users#new`.

src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -538,9 +539,12 @@ def import_available?
     #
     # @see Y2Users::InstUsersDialogHelper#user_to_validate
     #
-    # @return [Y2Users::User] root user if using entered password for it too; the new user otherwise
+    # @return [Y2Users::User] a root user when using same password for root; a normal user otherwise
     def user_to_validate
-      @use_pw_for_root ? root_user : user
+      username = @user_pw_for_root ? "root" : "user"
```
joseivanlopez, line 544: I would avoid to create a temporary user only for validations. This could lead to some troubles if we do not create the user rightly. I think the dialog can directly work with a user that is not attached to any config and then it can perform validations over such a user. The resulting user from the dialog could be attached/updated in the corresponding config once the dialog is accepted. And, if the dialog is discarded, then the user is simply gone.

joseivanlopez: As we have discussed, this can be addressed in a separate PR.

joseivanlopez (approved):
LGTM

PR #275, opened 2021-05-12
Fix autologin

joseivanlopez:
## Problem

After adapting the [*users_finish* client](https://github.com/yast/yast-users/pull/270) to use the new `Y2Users` classes, the autologin feature is not working anymore.

## Solution

Autologin config was not written when creating the users. Fixed.

## Testing

* Added unit tests
* Manually tested

Note: there still is a bug with the root password form, see #273.

### DIFF
--- a/test/users_finish_test.rb
+++ b/test/users_finish_test.rb
@@ -82,6 +82,9 @@ describe Yast::UsersFinishClient do
 
           allow(system_config).to receive(:merge).and_return(target_config)
 
+          # Mocking to avoid to write to the system
+          allow(Yast::Autologin).to receive(:Write)
+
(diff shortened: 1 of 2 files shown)


test/users_finish_test.rb
```diff
@@ -153,6 +162,12 @@
 
             subject.run
           end
+
+          it "does not write the auto-login config" do
+            expect(Yast::Autologin).to_not receive(:Write)
```
ancorgs, line 167: I wonder whether the correct thing to do is to not call to `Write`. Maybe we should actually call `User(false) && Write`to ensure it's disabled?

joseivanlopez: Good point!

ancorgs (approved):
LGTM

PR #279, opened 2021-05-18
After changing the users, invalidate the nscd cache and recreate the NIS database

ancorgs:
Still a WIP


ancorgs:
Closed in favor of #281

PR #280, opened 2021-05-18
Adapting Authorized keys

dgdavid:


src/lib/y2users/linux/writer.rb
```diff
@@ -216,6 +217,24 @@ def change_password(user, issues)
...
       end
 
+      # Writes authorized keys for given user
+      #
+      # @see Yast::Users::SSHAuthorizedKeyring#write_keys
+      #
+      # @param user [Y2User::User]
+      # @param issues [Y2Issues::List] a collection for adding issues if something goes wrong
+      def write_auth_keys(user, issues)
+        return unless user.home
```
joseivanlopez, line 227: NP: I would log this case: "Authorized keys cannot be written because the user has no home".

src/lib/y2users/linux/writer.rb
```diff
@@ -216,6 +217,24 @@ def change_password(user, issues)
...
+
+        Yast::Users::SSHAuthorizedKeyring.new(user.home, user.authorized_keys).write_keys
+      rescue Yast::Users::SSHAuthorizedKeyring::PathError => e
+        issues << Y2Issues::Issue.new(
+          # TRANSLATORS: %s is a placeholder for a username
+          format(_("Error writing authorized keys for '%s'"), user.name)
+        )
+        log.error("Error writing authorized keys for '#{user.name}' - #{e.message}")
+      end
+
```
joseivanlopez, line 237: NP: As I commented in another PR, this writer is growing and growing. I feel we need to restructure it with something like Action classes. But this is probably out of the scope of this PR.

src/lib/y2users/user.rb
```diff
@@ -86,6 +86,13 @@ class User < ConfigElement
     # @return [Password]
     attr_accessor :password
 
+    # Authorized keys
+    #
+    # TODO: use an specialized SSH public keys collection
```
joseivanlopez, line 91: Why is a collection needed?

src/lib/y2users/user.rb
```diff
@@ -154,7 +162,7 @@ def uid?
     def ==(other)
       return false unless other.is_a?(self.class)
 
-      [:name, :uid, :gid, :shell, :home, :gecos, :source, :password].all? do |a|
+      [:name, :uid, :gid, :shell, :home, :gecos, :source, :password, :authorized_keys].all? do |a|
```
joseivanlopez, line 165: I think that *authorized_keys* array has to be duplicated when coping a user (`#copy` method). Otherwise:

~~~
user1.authorized_keys = ["key1", "key2"]
user2 = user1.copy
user2.authorized_keys << "key3"
user1.authorized_keys #=> ["key1", "key2", "key3"]
~~~

joseivanlopez: Something like:

~~~
def copy(other)
  ...
  authorized_keys = other.authorized_keys.map(&:dup)
  ...
end
~~~

test/lib/y2users/linux/writer_test.rb
```diff
@@ -51,6 +51,39 @@
...
+    before do
+      allow(Yast::Execute).to receive(:on_target!)
+      allow(Yast::Users::SSHAuthorizedKeyring).to receive(:new).and_return(keyring)
+    end
+
+    RSpec.shared_examples "writing authorized keys" do
+      let(:home) { "/home/y2test" }
+
+      before do
+        current_user = config.users.find { |u| u.id == user.id }
```
joseivanlopez, line 65: NP: we can use the new API here:

~~~
current_user = config.users.by_id(user.id)
~~~

test/lib/y2users/linux/writer_test.rb
```diff
@@ -216,6 +245,15 @@
         writer.write
       end
 
+      context "whose authorized keys were edited" do
+        before do
+          current_user = config.users.find { |u| u.id == user.id }
```
joseivanlopez, line 250: The same here.

test/lib/y2users/user_test.rb
```diff
@@ -29,6 +29,12 @@
 
   include_examples "config element"
 
+  describe "#authorized_keys" do
+    it "returns a collection" do
```
joseivanlopez, line 33: NP: Maybe using the "collection" term here could be confusing.

src/lib/y2users/user.rb
dgdavid: Ah, good point. I overlooked it.

test/lib/y2users/linux/writer_test.rb
dgdavid: Right, I didn't realize it when rebasing changes

src/lib/y2users/user.rb
dgdavid: As discussed _offline_, because in the feature we might end up needing a collection to manage them when working in an installed system. But, as agreed, let's drop the TODO by now since for the current code base an Array is more than enough.

Thanks!

src/lib/y2users/linux/writer.rb
dgdavid: Think that at this point we don't know if there are keys to being written. That's the job of SSHAuthorizedKeyring#write_keys, which first _clean_ the old ones and then writes the new ones, if any. But it crash when _home_ is not defined (because a nil.to_s)

Perhaps it should do nothing if home is nil

src/lib/users/widgets/public_key_selector.rb
```diff
@@ -75,7 +78,8 @@ def handle(event)
 
       # @see CWM::AbstractWdiget
       def store
-        Yast::UsersSimple.SetRootPublicKey(value.to_s)
+        root_user.authorized_keys = [value.to_s]
+        Y2Users::UsersSimple::Writer.new(users_config).write
```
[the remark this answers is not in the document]
dgdavid: As agreed, let's address it as part of this PR

src/lib/y2users/linux/writer.rb
joseivanlopez: I expect that `user.authorized_keys == initial_user.authorized_keys` for the system users, so I think this method is not called for them.

dgdavid:
@joseivanlopez, 

I made changes agreed yesterday afternoon. Principally dropping the `InstUsersDialogHelper`, not loading a new reading of Yast::UsersSimple in each widget of the `isnt_root_first` dialog, and using a user copy to make changes until validation success.

Sadly, the approach of passing down the `root_user` to the widgets of `isnt_root_first` dialog and executing the `Y2Users:UsersSimple::Writer#write` in the `#next_handler` does not work. It is too early, the #store methods of widgets have been not triggered yet :confused::confused: and the user still untouched

Do you have any idea how to _fix it_? Do you know if there is a _hook_ to perform actions after all CWM widgets finished their work?

If not, I guess we should revert the latest changes and (except fixes here and there) for unblocking this PBI and address the dialog/widgets improvements in a separated PR.

src/lib/y2users/user_validator.rb
```diff
@@ -58,12 +58,6 @@ def issues
       err = Yast::UsersSimple.CheckFullname(user.full_name)
       add_fatal_issue(list, err, FULL_NAME_LOC)
 
-      return list unless user.password
-
-      user.password.issues.map do |issue|
-        list << issue
-      end
-
```
dgdavid, line 66: @ancorgs,

I believe this is a dead code, isn't it? Notice the dot between `password` and `issues` which is making the validator crash when validating a user with a password. I was thinking if you simply wanted to write `password_issues`, but if I remember correctly we are performing such validation separated on purpose.

Please, check it.

src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -321,22 +392,27 @@ def process_new_user_form
...
-    def attach_user(username, full_name: nil)
-      users_config.detach(user) if user
+    # Updates the users configuration with the given user
+    #
+    # @param user [Y2Users::User] the user for updating the configuration
+    def update_config(user)
+      main_user.name = user.name
+      main_user.gecos = user.gecos
+      main_user.password = user.password
+      root_user.password = user.password if @use_pw_for_root
```
dgdavid, line 413: I'm not so happy with the naming here (well, not sure if I'm actually happy with the inst_user_first after restructure it again). I tried to avoid using `tmp_user` or `form_user` in the previous method (#process_new_user_form) and end up choosing `main_user` for the user who is going to receive the updates after validations succeeded. Improvements are welcome, as long as we decide to go ahead with this change (see https://github.com/yast/yast-users/pull/280#issuecomment-845570702)

joseivanlopez:
> @joseivanlopez,
> 
> I made changes agreed yesterday afternoon. Principally dropping the `InstUsersDialogHelper`, not loading a new reading of Yast::UsersSimple in each widget of the `isnt_root_first` dialog, and using a user copy to make changes until validation success.
> 
> Sadly, the approach of passing down the `root_user` to the widgets of `isnt_root_first` dialog and executing the `Y2Users:UsersSimple::Writer#write` in the `#next_handler` does not work. It is too early, the #store methods of widgets have been not triggered yet confusedconfused and the user still untouched
> 
> Do you have any idea how to _fix it_? Do you know if there is a _hook_ to perform actions after all CWM widgets finished their work?

I didn't remember that "small" detail about the order of validations and next handler. In that case, I would go with something like we usually do in the partitioner:

* The dialog configures the user properly and nothing else.
* The caller of the dialog (i.e., the client) calls the writer in case that the dialog was accepted.

Maybe we need to pass the config object from the client to the dialog, or we can use the ConfigManager.

> 
> If not, I guess we should revert the latest changes and (except fixes here and there) for unblocking this PBI and address the dialog/widgets improvements in a separated PR.

IMHO, going back to the previous state is a bad idea. It was working because thanks to two different bugs (one in the dialog and another one in the writer) which were hiding the issue when writing the user.

> 
> EDIT: BTW, better to review commit by commit after your last review. Hint, start [here](https://github.com/yast/yast-users/pull/280/commits/50d86f59e4c2ac7161cad5105f925ed8dadff8b3).

src/lib/y2users/user_validator.rb
dgdavid: Ok, as discussed in a meeting, it was not dead but broken code. So, it has been restored and fixed. In addition, the ability to skip the password validations has been added through the 'skip_password` named param.

src/lib/y2users/user.rb
```diff
@@ -213,15 +219,17 @@ def system=(value)
...
     end
 
     # Validation errors
     #
+    # @param skip [Array<Symbol>] attributes that must be not checked. See {UserValidator#issues}
     # @return [Y2Issues::List]
-    def issues
-      UserValidator.new(self).issues
+    def issues(skip: [])
+      UserValidator.new(self).issues(skip: skip)
```
dgdavid, line 232: @ancorgs 

Now that I updated this method, I have the doubt if actually `User#issues` should be just `UserValidator#issues(skip: [:password])` directly, without params. I mean

```suggestion
    def issues
      UserValidator.new(self).issues(skip: [:password])
```

I.e, `User#issues` and `User#password_issues` as a shortcuts of `UserValidator#issues(skip: [:password])` and `UserValidator#issues(skip: [:name, :full_name])`

What do you think?

dgdavid: As agreed in a video call, let's simply pass down given args. Thanks

src/lib/y2users/parsers/passwd.rb
```diff
@@ -47,6 +48,7 @@ def parse(content)
           user.shell = values[PASSWD_MAPPING["shell"]]
           user.gecos = values[PASSWD_MAPPING["gecos"]].to_s.split(",")
           user.home =  values[PASSWD_MAPPING["home"]]
+          user.authorized_keys = Yast::Users::SSHAuthorizedKeyring.new(user.home).read_keys
```
joseivanlopez, line 51: Umm, the authorized keys is not a passwd field. I would move this to another place, maybe reader?

src/lib/y2users/password_helper.rb
```diff
@@ -0,0 +1,59 @@
...
+require "yast"
+require "y2users"
+require "y2users/users_simple"
+
+Yast.import "UI"
+Yast.import "Popup"
+Yast.import "Report"
+
+module Y2Users
+  # Mixin to hold shared helper methods related to the user password
```
joseivanlopez, line 29: Please, document that this helper is used by dialogs, because it takes cares of UI things like setting the focus, etc.

src/lib/y2users/user.rb
```diff
@@ -209,19 +215,24 @@ def system=(value)
...
     end
 
     # Validation errors
     #
+    # @see UserValidator#issues
+    #
     # @return [Y2Issues::List]
-    def issues
-      UserValidator.new(self).issues
+    def issues(*args)
```
joseivanlopez, line 228: Please, document the *args* parameter.

test/lib/y2users/users_simple/writer_test.rb
```diff
@@ -65,10 +67,13 @@ def user_simple(name)
       context "when root has no password" do
         let(:root_password) { nil }
 
-        it "does not store a password for root" do
-          subject.write
+        # NOTE: for Yast::UsersSimple emtpy string means no password. Thus, not
```
joseivanlopez, line 70: Typo: emtpy

src/lib/y2users/linux/local_reader.rb
```diff
@@ -79,6 +83,14 @@ def read_passwords(config)
           user.password = password
         end
       end
+
+      def read_authorized_keys(config)
```
joseivanlopez, line 87: It is ok for now, but I think that `Linux::Reader` and `Linux::LocalReader` deserve a `Linux::BaseReader` class to allocate shared logic like this for reading the authorized keys.

joseivanlopez (approved):
LGTM

src/lib/y2users/linux/local_reader.rb
dgdavid: Yes, I thought exactly the same few minutes ago, when moving the read of authorized_keys to those readers.

PR #282, opened 2021-05-20
Improve readers API

joseivanlopez:
## Problem

Currently, readers offer a `#read_to` method which receives a config argument. But having a config parameter is not needed and it could cause misunderstandings.  For example, what happens with the elements that the given config already has? Should the reader draw its elements off before reading? Should the current elements be updated somehow?

## Solution

Improve the API by removing the config parameter. Now the `#read` method always generates a new config.


imobachgs (approved):
IMHO, that's much better now. And it is analogous to what we have in yast2-network.

PR #283, opened 2021-05-21
Improve comparisson

joseivanlopez:
In order to


src/lib/y2users/equatable.rb
```diff
@@ -0,0 +1,45 @@
...
+  module ClassMethods
+    attr_reader :eql_attrs
+
+    def eql_attr(*names)
+      @eql_attrs ||= []
+      @eql_attrs += names
+    end
+  end
+
+  # Other strategies can be used in order to generate the hash key, like creating a Hash object and
```
teclator, line 34: Originally I was thinking on creating a Hash with the attributes / values calling hash over it but we could also call hash over a sequence

https://ruby-doc.org/core-2.6.3/Enumerator/ArithmeticSequence.html#method-i-hash

Similar or basically what the Equatable gem does.

https://github.com/piotrmurach/equatable/blob/master/lib/equatable.rb#L102

Did you get inspired by that gem already?.

joseivanlopez: I did not know it :)

teclator: Me neither ;)

test/lib/y2users/equatable_test.rb
```diff
@@ -0,0 +1,147 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require_relative "test_helper"
+require "y2users/equatable"
+
+describe Y2Users::Equatable do
```
teclator, line 25: In general LGTM.

np: as it implements the **hash** method I miss a unit test covering it and also for the **#==** one which could be a shared example for both **eql?** and **#==**.

joseivanlopez: The `#hash` method in indirect tested through `#eql?`. Its test would be basically the same. And I am not sure for `#==`, they are alias. But ok, I can run the very same tests for it.

teclator: Yep, both are indirect tested, it would be more about documentation than other thing but it looks good to me as it is.

src/lib/y2users/user.rb
```diff
@@ -145,22 +150,6 @@ def uid?
       !(uid.nil? || uid == "")
     end
 
-    # Whether two users are equal
-    #
-    # Only relevant attributes are compared. For example, the config in which the user is attached
```
ancorgs, line 150: I wouldn't remove the comment mentioning than the internal id and the config were intentionally left out.

imobachgs (approved):

PR #284, opened 2021-05-25
Manage useradd configuration

ancorgs:
This adds management of the useradd configuration, which is important for several reasons:

- The AutoYaST profile can contain a `<user_defaults>` section used to configure useradd
- The YaST UI for managing users also contain a tab for configuring useradd
- The useradd default values must be used in the YaST UI to pre-fill several fields

See also https://github.com/yast/yast-yast2/pull/1166 and https://trello.com/c/vhM38X7J/2463-3-manage-useradd-configuration

### DIFF
--- a/src/lib/y2users/config.rb
+++ b/src/lib/y2users/config.rb
@@ -102,6 +102,7 @@ module Y2Users
 
       config = self.class.new
       config.attach(elements)
+      config.useradd_config = useradd_config.dup
       config
     end
 
@@ -124,6 +125,11 @@ module Y2Users
 
       self
     end
+
+    # Useradd configuration to be applied to the system before creating the users
+    #
+    # @return [UseraddConfig, nil] nil if the configuration is unknown
+    attr_accessor :useradd_config
 
   private
 
--- a/src/lib/y2users/config.rb
+++ b/src/lib/y2users/config.rb
@@ -102,7 +102,7 @@ module Y2Users
 
       config = self.class.new
       config.attach(elements)
-      config.useradd_config = useradd_config.dup
+      config.useradd = useradd.dup
       config
     end
 
@@ -129,7 +129,7 @@ module Y2Users
     # Useradd configuration to be applied to the system before creating the users
     #
     # @return [UseraddConfig, nil] nil if the configuration is unknown
-    attr_accessor :useradd_config
+    attr_accessor :useradd
 
   private
 
--- a/src/lib/y2users/linux/useradd_config_reader.rb
+++ b/src/lib/y2users/linux/useradd_config_reader.rb
@@ -1,0 +1,88 @@
+# Copyright (c) [2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "yast2/execute"
+require "y2users/useradd_config"
+
+module Y2Users
+  module Linux
+    # Reads the useradd configuration from the current system
+    class UseraddConfigReader
+      include Yast::Logger
+      Yast.import "ShadowConfig"
+
+      # Creates a {UseraddConfig} object with the system configuration
+      #
+      # @return [UseraddConfig]
+      def read
+        attrs = read_shadow_config.merge(read_useradd)
+        UseraddConfig.new(attrs)
+      end
+
+    private
+
+      # Command for reading the useradd default values
+      USERADD = "/usr/sbin/useradd".freeze
+      private_constant :USERADD
+
+      # Mapping between useradd keys and {UseraddConfig} attributes
+      USERADD_ATTRS = {
+        group:             "GROUP",
+        home:              "HOME",
+        inactivity_period: "INACTIVE",
+        expiration:        "EXPIRE",
+        shell:             "SHELL",
+        skel:              "SKEL",
+        usrskel:           "USRSKEL",
+        create_mail_spool: "CREATE_MAIL_SPOOL"
+      }.freeze
+      private_constant :USERADD_ATTRS
+
+      # Values from "useradd -D"
+      #
+      # @return [Hash]
+      def read_useradd
+        output = Yast::Execute.on_target!(USERADD, "-D", stdout: :capture)
+        values = output.lines.map(&:strip).map { |line| line.split("=", -1) }.to_h
+        USERADD_ATTRS.keys.map { |attr| [attr, useradd_value(values, attr)] }.to_h
+      rescue Cheetah::ExecutionFailed => e
+        log.warn("Failed to read useradd default values' - #{e.message}")
+        {}
+      end
+
+      # @see #read_useradd
+      def useradd_value(values, attr)
+        value = values[USERADD_ATTRS[attr]]
+        return value unless [:inactivity_period, :create_mail_spool].include?(attr)
+
+        return value.to_i if attr == :inactivity_period
+
+        value == "yes"
+      end
+
+      # Values from login.defs
+      #
+      # @return [Hash]
+      def read_shadow_config
+        { umask: Yast::ShadowConfig.fetch(:umask) }
+      end
+    end
+  end
+end
--- a/src/lib/y2users/linux/useradd_config_reader.rb
+++ b/src/lib/y2users/linux/useradd_config_reader.rb
@@ -70,11 +70,15 @@ module Y2Users
       # @see #read_useradd
       def useradd_value(values, attr)
         value = values[USERADD_ATTRS[attr]]
-        return value unless [:inactivity_period, :create_mail_spool].include?(attr)
 
-        return value.to_i if attr == :inactivity_period
-
-        value == "yes"
+        case attr
+        when :inactivity_period
+          value.to_i
+        when :create_mail_spool
+          value == "yes"
+        else
+          value
+        end
       end
 
       # Values from login.defs
--- a/src/lib/y2users/linux/useradd_config_writer.rb
+++ b/src/lib/y2users/linux/useradd_config_writer.rb
@@ -31,8 +31,8 @@ module Y2Users
 
       # Constructor
       #
-      # @param config [Y2User::Config] see #config
-      # @param initial_config [Y2User::Config] see #initial_config
+      # @param config [Config] see #config
+      # @param initial_config [Config] see #initial_config
       def initialize(config, initial_config)
         textdomain "y2users"
 
@@ -52,25 +52,25 @@ module Y2Users
 
       # General configuration object containing the useradd configuration to apply to the system
       #
-      # @return [Y2User::Config]
+      # @return [Config]
       attr_reader :config
 
-      # Initial state of the system (usually a Y2User::Config.system in a running system) that will
+      # Initial state of the system (usually a Y2Users::Config.system in a running system) that will
       # be compared with {#config} to know what changes need to be performed.
       #
-      # @return [Y2User::Config]
+      # @return [Config]
       attr_reader :initial_config
 
       # Object containing the useradd configuration
       #
-      # @return [Y2User::UseraddConfig]
+      # @return [UseraddConfig]
       def useradd_config
         config&.useradd_config
       end
 
       # Object containing the useradd configuration fron the initial state
       #
-      # @return [Y2User::UseraddConfig]
+      # @return [UseraddConfig]
       def initial_useradd_config
         initial_config&.useradd_config
       end
--- a/src/lib/y2users/linux/useradd_config_writer.rb
+++ b/src/lib/y2users/linux/useradd_config_writer.rb
@@ -65,14 +65,14 @@ module Y2Users
       #
       # @return [UseraddConfig]
       def useradd_config
-        config&.useradd_config
+        config&.useradd
       end
 
       # Object containing the useradd configuration fron the initial state
       #
       # @return [UseraddConfig]
       def initial_useradd_config
-        initial_config&.useradd_config
+        initial_config&.useradd
       end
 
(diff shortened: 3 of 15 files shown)


src/lib/y2users/linux/useradd_config_writer.rb
```diff
@@ -0,0 +1,174 @@
...
+  module Linux
+    # Configures useradd in the target system according to a given {UseraddConfig} object
+    class UseraddConfigWriter
+      include Yast::I18n
+      include Yast::Logger
+      Yast.import "ShadowConfig"
+
+      # Constructor
+      #
+      # @param config [Y2User::Config] see #config
```
dgdavid, line 34: Wrong namespace in documentation, missing `s`. Same happens several times,  please check.

```suggestion
      # @param config [Y2Users::Config] see #config
```

test/lib/y2users/useradd_config_test.rb
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env rspec
```
dgdavid, line 1: Please, include this new file in _test/Makefile.am_

dgdavid (approved):
LGTM, nice work.

Just two minor comments.

src/lib/y2users/config.rb
```diff
@@ -125,6 +126,11 @@ def merge!(other)
       self
     end
 
+    # Useradd configuration to be applied to the system before creating the users
+    #
+    # @return [UseraddConfig, nil] nil if the configuration is unknown
+    attr_accessor :useradd_config
```
joseivanlopez, line 132: I think we need to copy/clone this object when doing a deep copy with `#copy` method.

src/lib/y2users/linux/useradd_config_reader.rb
```diff
@@ -0,0 +1,88 @@
...
+
+      # Values from "useradd -D"
+      #
+      # @return [Hash]
+      def read_useradd
+        output = Yast::Execute.on_target!(USERADD, "-D", stdout: :capture)
+        values = output.lines.map(&:strip).map { |line| line.split("=", -1) }.to_h
+        USERADD_ATTRS.keys.map { |attr| [attr, useradd_value(values, attr)] }.to_h
+      rescue Cheetah::ExecutionFailed => e
+        log.warn("Failed to read useradd default values' - #{e.message}")
```
jreidinger, line 66: no support for issues? I do not like much this behavior as not set default is same as if it failed to read.

src/lib/y2users/linux/useradd_config_reader.rb
```diff
@@ -0,0 +1,88 @@
...
+      end
+
+      # @see #read_useradd
+      def useradd_value(values, attr)
+        value = values[USERADD_ATTRS[attr]]
+        return value unless [:inactivity_period, :create_mail_spool].include?(attr)
+
+        return value.to_i if attr == :inactivity_period
+
+        value == "yes"
```
jreidinger, line 77: so this is only for `:create_mail_spool`. It is quite hidden. What about using case
```
case attr
when :inactivity_period then value.to_i
when :create_mail_spool then value == "yes"
else value
end

File: src/lib/y2users/linux/useradd_config_writer.rb [hunk 6]
```diff
@@ -0,0 +1,174 @@
...
+      # @param initial_config [Y2User::Config] see #initial_config
+      def initialize(config, initial_config)
+        textdomain "y2users"
```
jreidinger, line 37: wrong textdomain

src/lib/y2users/config.rb
joseivanlopez: And what happens when merging configs?

src/lib/y2users/linux/useradd_config_writer.rb
```diff
@@ -0,0 +1,174 @@
...
+      # @param issues [Y2Issues::List]
+      def configure_useradd_attr(attr, arg, issues)
+        value = value(attr)
+        return if value.nil?
+
+        Yast::Execute.on_target!(USERADD, "-D", arg, value)
+      rescue Cheetah::ExecutionFailed => e
+        issues << Y2Issues::Issue.new(
+          # TRANSLATORS: %s is the name of one of the useradd default values, like 'HOME' or 'GROUP'
+          format(_("Something went wrong writing the useradd default for '%s'"), arg)
```
jreidinger, line 168: I think this issues message is something we need to address. It is quite useless now. You just know that something went wrong, but not if it is written or not, what is error message, etc.

src/lib/y2users/linux/writer.rb
```diff
@@ -100,6 +101,8 @@ def write
         issues = Y2Issues::List.new
 
         add_groups(issues)
+        # Configure useradd after creating the groups, to make sure the default group exists already
```
jreidinger, line 104: So add_groups reads the gid of newly created group? Otherwise it won't work.

src/lib/y2users/useradd_config.rb
```diff
@@ -0,0 +1,186 @@
...
+      # Only attributes in this list have a public setter
+      #
+      # @return [Array<Symbol>]
+      def writable_attributes
+        WRITABLE_ATTRIBUTES.dup
+      end
+
+    private
+
+      # Internal class method to automatically define the setters for all writable attributes
```
jreidinger, line 72: I worry this way documentation will miss info that ttrs are writable and setter exists at all.

src/lib/y2users/config.rb
```diff
@@ -102,6 +102,7 @@ def copy
 
       config = self.class.new
       config.attach(elements)
+      config.useradd_config = useradd_config.dup
```
joseivanlopez, line 105: What happens when merging configs? How is this object merged?

src/lib/y2users/linux/writer.rb
```diff
@@ -149,6 +152,13 @@ def write
       # @return [Y2User::Config]
       attr_reader :initial_config
 
+      # Writes the useradd configuration to the system
+      #
+      # @param issues [Y2Issues::List]
+      def write_useradd_config(issues)
+        UseraddConfigWriter.new(config, initial_config).write(issues)
```
joseivanlopez, line 159: NP: I would prefer to generate a list of issues (like other writers)  instead of passing the issues list.

src/lib/y2users/linux/useradd_config_writer.rb
dgdavid: Right, good catch!

src/lib/y2users/linux/writer.rb
ancorgs: I did it that way because this writer is called from the more general Linux writer and so far we have followed the approach of creating an empty list at the beginning of the whole process and then passing that list down for each step to accumulate issues in that original list. It's simply not so obvious now because all the steps of the reader are (wrongly, IMHO) implemented via methods, all of them mixed into the same big class, instead of separate smaller classes (like this one).

So the change you suggest would also be inconsistent from some point of view.

I would agree if the current implementation of `Linux::Writer#read` would be something like this:

```ruby
issues = add_groups
issues += write_useradd_config # This instantiates a UseraddConfigWriter internally
issues += add_users
issues += edit_users

issues
```

But there is no method `#+` or `#merge` or similar for `Y2Issues::List`. We could, of course, go all the way down and implement such method and then stop passing the list to the new partial writer AND also to all the methods like `add_groups`. But fixing just the `write_useradd_config` step because is the only one that uses a separate class looks like barking to the wrong tree to me.

src/lib/y2users/useradd_config.rb
ancorgs: Maybe. Let me lookup whether there is a solution. So far, that very well resolved in Crystal with macros :-)

src/lib/y2users/linux/writer.rb
ancorgs: Why? Useradd will use that group by default in those cases in which the Ruby object does not provide a group. So this is only relevant for useradd itself, who reads the users from the system and has no visibility on the list of groups we have in our Ruby objects. It's somehow irrelevant for our code.

src/lib/y2users/linux/useradd_config_reader.rb
ancorgs: Well, this is called from the more general Linux reader, which does not report issues right now. So no matter if the `UseraddConfigReader` collects issues, they will be discarded by the caller anyways.

None of the readers use the issues mechanism right now. We can open a separate card to reconsider that (including deciding when to show those errors, etc.). But I think it's out of scope here.

imobachgs: JFYI, I plan to add an issue when the username is not specified in AutoYaST.

src/lib/y2users/linux/writer.rb
joseivanlopez: Yes, it has been pointed several times that it would be nice to split this writer into smaller classes. And I don't get why we decided to pass the *issues* parameter all over the place instead of using an object attribute, something like:

~~~
def write
  @issues = Y2Issues::List.new
  
  add_groups
  write_useradd_config
  ...

  @issues
end
~~~

But in general I agree that ideally we should have something like `Y2Issues::List#concat`. Anyway, IMHO, passing the issues parameter to the new writer is a movement in the opposite way we would like to go. But it is only a detail, nothing super important from my point of view ;)

dgdavid: > And I don't get why we decided to pass the _issues_ parameter all over the place instead of using an object attribute,

I did it at the very beginning (to use an object attribute), but it was suggested to use a param instead (for the sake of simplicity, I guess)

src/lib/y2users/linux/useradd_config_writer.rb
ancorgs: This was copied&pasted from `Linux::Writer` which, apart from being wrong, redundantly specifies the namespace everywhere. This proves that's not a great idea, so I will remove the excess of namespaces in both files.

src/lib/y2users/linux/useradd_config_writer.rb
ancorgs: Giving a more accurate message would likely imply parsing the stderr of "useradd -D" and similar stuff that I consider overkill taking into consideration the details can be looked up in the YaST logs. But I'm open to suggestions here.

src/lib/y2users/linux/useradd_config_writer.rb
ancorgs: Fixed now.

test/lib/y2users/useradd_config_test.rb
ancorgs: Done

src/lib/y2users/linux/useradd_config_reader.rb
ancorgs: Done.

src/lib/y2users/linux/useradd_config_writer.rb
ancorgs: Fixed.

src/lib/y2users/useradd_config.rb
ancorgs: I changed the strategy a bit and now it's properly documented in the yardoc output.

src/lib/y2users/config.rb
ancorgs: I haven't though about it. I wonder whether it's actually relevant. Can you describe the "use case"?

src/lib/y2users/linux/writer.rb
ancorgs: Since this is extracted to a new card (which would imply the creation of `Y2Issues::List#concat`), I'm leaving it as it is for the time being.

https://trello.com/c/yVrdlyCx/4747-refactor-linuxwriter

src/lib/y2users/linux/useradd_config_reader.rb
ancorgs: > JFYI, I plan to add an issue when the username is not specified in AutoYaST.

I guess the AutoYaST reader is a special case, since there is indeed a step to show the issues found while reading the profile. But we have no such thing in general for reporting the issues that happened while analyzing the system. Not yet, at least.

If we add it, we need to discuss when and how to present such issues to the user in the current workflow.

ancorgs: I just added a comment to https://trello.com/c/08t84wmL/4716-a-better-way-for-reporting-issues-at-the-end-of-installation

src/lib/y2users/config.rb
joseivanlopez: Maybe when using AutoYaST in a running system. The system config will be merged with the config from the AutoYaST profile, and both configs will have their own useradd conf.

ancorgs: May be. But that's something to be solved at https://trello.com/c/QU2gDrTK/2486-3-autoyast-userdefaults-section (already planned for this sprint)

joseivanlopez (approved):
LGTM

src/lib/y2users/linux/useradd_config_reader.rb
jreidinger: fine for me

src/lib/y2users/linux/useradd_config_writer.rb
jreidinger: ok, I already discuss it with imo that it would be nice if there is something like details where we can just pass untranslated stderr.

src/lib/y2users/linux/writer.rb
jreidinger: OK, lets consider this scenario:

- default group specified by group name in autoyast profile
- group does not have gid in autoyast profile
- group does not exist yet on system

So what we need to do in such case is to create group, read gid it gets assigned and use it in this useradd_config...or does it support also group name instead of gid as argument?

ancorgs: Yes. Useradd supports both group name and gid for the GROUP configuration key. Actually, what AutoYaST does is copying the value from the profile as-is (be it a number or a name) to the useradd configuration. No need to do any conversion.

jreidinger: ok, then it is fine for me. Just this comment confuse me as it looks like it is mandatory to have group existing before configuring useradd, which I am not sure if it is case. I think it is just needed before calling useradd when creating user, not?

ancorgs: Yes, I can rephrase it a bit

ancorgs: Although it's actually true that the group must exists before configuring useradd. Otherwise useradd will complain like this:

`
$ useradd -D --gid not_a_group
useradd: group 'not_a_group' does not exist
`

jreidinger (approved):

PR #285, opened 2021-05-26
Add classes to handle the AutoYaST profile

imobachgs:
Add a set of classes to process the AutoYaST profile and adapt the `Autoinst::Reader` to use those classes.


teclator (approved):
LGTM

src/lib/y2users/autoinst/reader.rb
```diff
@@ -98,12 +103,13 @@ def create_password(user)
       # @param user [Hash] a user representation in the format used by UsersSimple
       # @return [String]
       def shadow_string(user)
-        pwd_settings = user["password_settings"] || {}
-        other_attrs = SORTED_SHADOW_ATTRS.map do |attr|
-          pwd_settings[attr] || ""
+        other_attrs = if user.password_settings
+          SORTED_SHADOW_ATTRS.map { |a| user.password_settings.send(a) || "" }
```
joseivanlopez, line 107: I wonder if we have here the very same problem that we are addressing with *UsersSimple*. I mean, there is no way to distinguish nil and "" values. For example, a nil "min" means that the value is unknown, so the writer will no keep it as it is. But "" value means that there is no minimum password age.

imobachgs: This method returns a String representing the entry in the shadow file, AFAIU (I did not implement this part). And given that the values in `other_attrs` are joined using ':', I guess the `||` is not needed at all. However, I do not think the difference between nil and "" in this context is relevant.

test/lib/y2users/autoinst_profile/user_section_test.rb
```diff
@@ -0,0 +1,65 @@
...
+        expect(section.authorized_keys).to eq([])
+      end
+    end
+
+    it "sets password settings" do
+      section = described_class.new_from_hashes("password_settings" => { "warn" => 60 })
+      expect(section.password_settings.warn).to eq(60)
+    end
+
+    context "when password settings" do
```
dgdavid, line 58: NP: when password settings is not present

src/lib/y2users/autoinst/reader.rb
imobachgs: We have discussed the issue "offline", as I did not understand the problem. It seems that this approach of ensembling the shadow line does not work when you need to distinguish nil and "". So in UsersSimple, we are following a different approach. We should do the same in another PR, following @joseivanlopez's approach.

Thanks for clarifying!

joseivanlopez (approved):
LGTM

PR #286, opened 2021-05-28
Adds a Linux::BaseReader

dgdavid:
For allocating shared logic between `Linux::LocalReader` and `Linux::Reader`, as commented in https://github.com/yast/yast-users/pull/280#discussion_r640739151


src/lib/y2users/linux/base_reader.rb
```diff
@@ -0,0 +1,114 @@
...
+      # @return [Config]
+      def read
+        elements = read_users + read_groups
+
+        config = Config.new.attach(elements)
+
+        # read passwords after user, as user has to exist in advance
+        read_passwords(config)
+
+        # read authorized keys
```
ancorgs, line 44: Do we need a comment saying "read authorized keys" to explain the line of code `read_authorized_keys(c)`?

dgdavid: Not at all

ancorgs (approved):
LGTM

PR #287, opened 2021-05-28
Fix aging issues

joseivanlopez:
## Problem

When the users information is written to *UsersSimple* module and then it is recovered again, it might result in wrong values for some of the password attributes like the "last change date". When this happens, the user could be forced to change his/her password in the first login attempt after the installation.

## Solution

Use `nil` value to indicate that a password attribute is unknown. When an attribute is `nil`, the linux writer will not pass the attribute to the *chage* commad, keeping the default values set by *chpasswd*. 

Moreover, the users simple reader is now able to keep the `nil` values that comes from the `UsersSimple` module.

## Testing

* Added unit tests.
* Manually tested.

### DIFF
--- a/src/lib/y2users/password.rb
+++ b/src/lib/y2users/password.rb
@@ -291,6 +291,9 @@ module Y2Users
     end
 
     # Disables the account expiration
+    #
+    # @note There is not a counterpart method called 'enable'. To activate the account expiration,
+    #   use {#date=} to assign an expiration date.
     def disable
       @content = ""
     end
--- a/src/lib/y2users/password.rb
+++ b/src/lib/y2users/password.rb
@@ -18,7 +18,7 @@
 # find current contact information at www.suse.com.
 
 require "yast2/secret_attributes"
-require "y2users/shadow_date_helper"
+require "y2users/shadow_date"
 require "yast2/equatable"
 
 module Y2Users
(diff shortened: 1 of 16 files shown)


dgdavid:
I have tested changes manually and (I think) now it works as expected

```
localhost:~ # chage -l dgdavid
Last password change					: may 28, 2021
Password expires					: never
Password inactive					: never
Account expires						: never
Minimum number of days between password change		: 0
Maximum number of days between password change		: 99999
Number of days of warning before password expires	: 7


localhost:~ # chage -l root
Last password change					: may 28, 2021
Password expires					: never
Password inactive					: never
Account expires						: never
Minimum number of days between password change		: -1
Maximum number of days between password change		: -1
Number of days of warning before password expires	: -1
```

src/lib/y2users/password.rb
```diff
@@ -222,21 +259,60 @@ def force_change
...
+  # Represents the account expiration field of the shadow file
+  class AccountExpiration < ShadowDateField
+    # Whether the account has an expiration date
+    #
+    # @return [Boolean]
+    def expire?
+      !content.empty?
+    end
+
+    # Disables the account expiration
```
ancorgs, line 293: Maybe we should also explain here that there is not an `#enable` counterpart and that `date=` should be used instead

ancorgs (approved):
LGTM

ancorgs (approved):
Looks even better with the improved doc.

PR #289, opened 2021-05-31
Manage login from config

joseivanlopez:
## Problem

Both interactive installation and AutoYaST take care of configuring the Autologin feature. In the interactive installation that's done through some checkboxes. And in AutoYaST that corresponds to the section *<login_settings>*. In both cases, configuring the *Autologin* module is needed and then it must be written to the system. 

* (Internal) https://trello.com/c/ncBZyAjF/2482-3-autologin-support

## Solution

Integrate *login* configuration in the `Y2Users::Config` object. Now, readers are able to add the login config to the configuration object. Then, writers write the autologin config properly.

* Requires https://github.com/yast/yast-yast2/pull/1170.

Bonus: some fixes for autoinst reader and linux writer. 

## Testing

* Added unit tests.
* Manually tested


src/lib/y2users/config.rb
```diff
@@ -170,6 +184,9 @@ def detach_element(element)
       collection_for(element).delete(element.id) if exist
 
       element.assign_config(nil)
+
+      # Clean up the autologin user if the detached user was set for autologin
+      login.autologin_user = nil if login? && login.autologin_user&.is?(element)
```
jreidinger, line 189: it is fine for me as long as it is documented in method about this side effect. Not sure if there is real use case for other behavior?

joseivanlopez: No idea about other possible use cases. I think this would be the expected result, I will document it.

src/lib/y2users/config.rb
```diff
@@ -67,6 +73,13 @@ def groups
       groups_collection.dup.freeze
     end
 
+    # Whether the login is configured
+    #
+    # @return [Boolean]
+    def login?
+      !login.nil?
```
jreidinger, line 80: To be honest I do not like much that spread of nil everywhere for not configured stuff when it can be done in object itself. See e.g. collections there. Why not have it in LoginConfig as attribute like `configured?` so here you can use `login.configured?`

src/lib/y2users/linux/writer.rb
```diff
@@ -225,6 +228,21 @@ def add_group(group, issues)
         log.error("Error creating group '#{group.name}' - #{e.message}")
       end
 
+      # Applies changes for the login settings
+      def edit_login(_issues)
+        return unless config.login?
+
+        # resetting Autologin settings
+        Yast::Autologin.Disable
```
jreidinger, line 236: so if it is not configured, it is not disabled?

src/lib/y2users/linux/writer.rb
```diff
@@ -225,6 +228,21 @@ def add_group(group, issues)
...
+
+        Yast::Autologin.user = config.login.autologin_user&.name.to_s
```
jreidinger, line 238: how it behave if user is nil or empty? it writes there empty string? Is difference if Use(false) is set and if user is set to empty string?

src/lib/y2users/users_simple/reader.rb
```diff
@@ -99,7 +120,9 @@ def attr_value(attr, user)
       # @param user [Hash] a user representation in the format used by UsersSimple
       # @return [Password]
       def create_password(user)
-        new_password(user["userPassword"], user["encrypted"]).tap do |password|
+        create_method = user["encrypted"] ? :create_encrypted : :create_plain
+
+        Password.send(create_method, user["userPassword"]).tap do |password|
```
jreidinger, line 125: at least use `public_send` other way could be to use passwordValue class and here just call `klass.new(user["userPassword"])`

src/lib/y2users/config.rb
joseivanlopez: Well, having nil value was the agreed decision to indicate that a value is unknown, see https://github.com/yast/yast-users/pull/287. In general, the writers will perform no action for nil attributes. Here, `#login` is following the same approach, so a `nil` login config means that it is unknown, so it should not persisted to the system. If we want to explicitly disable autologin, then we need to configure the login config properly. The same approach is also followed by `#useradd` here https://github.com/yast/yast-users/pull/284.

src/lib/y2users/linux/writer.rb
joseivanlopez: Exactly, if the login is unknown, then the writer should do nothing.

src/lib/y2users/linux/writer.rb
joseivanlopez: With `Autologin.Use(false)` the method `Autologin.Write` does nothing. Note that Autologin allows to configure both, the *autologin* user and also the *password less login* property. With `Autologin.Use(false)` none of these are written. For disabling autologin, we need to set  `Autologin.Use(true)` and set an empty string as autologin username.

src/lib/y2users/config.rb
```diff
@@ -67,6 +73,13 @@ def groups
       groups_collection.dup.freeze
     end
 
+    # Whether the login is configured
```
jreidinger, line 76: I think this is root of my confusion. I think something like `Whether the login configuration is defined` make it clear that it is not about configure login, but if that configuration is defined or unknown.

src/lib/y2users/config.rb
```diff
@@ -40,13 +40,19 @@ module Y2Users
   #   config2.detach(user)
   #   config2.users #=> UsersCollection<[user2]>
   class Config
+    # Login configuration, see {LoginConfig}
+    #
+    # @return [LoginConfig]
```
jreidinger, line 45: please document also nil case and maybe `@see login?` can be also useful.

jreidinger (approved):

PR #290, opened 2021-06-01
Modify user attributes (home, gecos, shell and groups)

imobachgs:
Bring back be876b7d0cbc28382f6f92c818a645788a1ac043 which added support for `usermod` and was wrongly removed by me when adapting the original PR.

### DIFF
--- a/src/lib/y2users/linux/writer.rb
+++ b/src/lib/y2users/linux/writer.rb
@@ -360,6 +360,8 @@ module Y2Users
       # @param old_user [Y2Users::User] Original user
       # @return [Array<String>] usermod options
       # rubocop:disable Metrics/CyclomaticComplexity
+      # rubocop:disable Metrics/AbcSize
+      # rubocop:disable Metrics/PerceivedComplexity
       def usermod_options(new_user, old_user)
         args = []
         args << "--gid" << new_user.gid if new_user.gid != old_user.gid && new_user.gid
@@ -375,6 +377,8 @@ module Y2Users
         args
       end
       # rubocop:enable Metrics/CyclomaticComplexity
+      # rubocop:enable Metrics/AbcSize
+      # rubocop:enable Metrics/PerceivedComplexity
 
       def different_groups?(user1, user2)
         user1.groups(with_primary: false).sort != user2.groups(with_primary: false).sort
--- a/src/lib/y2users/linux/writer.rb
+++ b/src/lib/y2users/linux/writer.rb
@@ -328,7 +328,7 @@ module Y2Users
       # @return [Array<String>] usermod options
       def usermod_options(new_user, old_user)
         args = []
-        args << "--comment" << new_user.gecos if new_user.gecos != old_user.gecos && new_user.gecos
+        args << "--comment" << new_user.gecos.join(",") if new_user.gecos != old_user.gecos
         if new_user.home != old_user.home && new_user.home
           args << "--home" << new_user.home << "--move-home"
         end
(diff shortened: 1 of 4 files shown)


src/lib/y2users/linux/writer.rb
```diff
@@ -271,6 +273,30 @@ def set_password_attributes(user, issues)
...
+          !new_user.public_send(attr).nil? &&
+            (new_user.public_send(attr) != old_user.public_send(attr))
+        end
+
+        Yast::Execute.on_target!(USERMOD, *usermod_options(new_user, old_user)) if usermod_changes
+        change_password(new_user, issues) if old_user.password != new_user.password
+        write_auth_keys(new_user, issues) if old_user.authorized_keys != new_user.authorized_keys
+      rescue Cheetah::ExecutionFailed => e
+        issues << Y2Issues::Issue.new(
+          format(_("The user '%{username}' could not be modified"), username: new_user.name)
```
jreidinger, line 295: Not related to this PR, but it is quite useless message without details. What about support for details in issues and add to that stderr where is usually useful details why it failed?

jreidinger:
sadly what is also missing and was removed originally is support for groups - https://github.com/yast/yast-users/pull/272/commits/913cb87223185c8a1ef4c4562d8299c971cbc15c and https://github.com/yast/yast-users/pull/272/commits/904a6c7a5a55715077815cf51355497d26ecfd99 it is needed as autoyast support groups and additional groups

imobachgs:
> sadly what is also missing and was removed originally is support for groups - [913cb87](https://github.com/yast/yast-users/commit/913cb87223185c8a1ef4c4562d8299c971cbc15c) and [904a6c7](https://github.com/yast/yast-users/commit/904a6c7a5a55715077815cf51355497d26ecfd99) it is needed as autoyast support groups and additional groups

Well, it is not exactly missing. As it is only supported by AutoYaST, we decided to move that logic to the `autoinst` reader. Actually, even the `groups` key was removed from `useradd`. I have a task in my todo list related to that.

imobachgs:
> > sadly what is also missing and was removed originally is support for groups - [913cb87](https://github.com/yast/yast-users/commit/913cb87223185c8a1ef4c4562d8299c971cbc15c) and [904a6c7](https://github.com/yast/yast-users/commit/904a6c7a5a55715077815cf51355497d26ecfd99) it is needed as autoyast support groups and additional groups
> 
> Well, it is not exactly missing. As it is only supported by AutoYaST, we decided to move that logic to the `autoinst` reader. Actually, even the `groups` key was removed from `useradd`. I have a task in my todo list related to that.

mmm perhaps I missread your comment. Anyway, I will introduce it in a different PR.

src/lib/y2users/linux/writer.rb
imobachgs: Yes, it is true. However, issues do not contain "details", so it is perhaps something to improve. Actually, you can find the details in the logs.

test/lib/y2users/config_merger_test.rb
```diff
@@ -194,10 +194,10 @@ def lhs_group(name)
           expect(lhs.groups.map(&:name)).to contain_exactly("test1", "test2", "test3")
         end
 
-        it "updates the lhs group with the data from the corresponding rhs" do
+        it "updates the lhs group with the data from the corresponding rhs group except gid" do
           subject.merge
 
-          expect(lhs_group("test1")).to eq(group1)
+          expect(lhs_group("test1").gid).to eq(100)
```
jreidinger, line 200: this is probably left over from my commit as it should not change merger behavior, so test should fail.

jreidinger (approved):

PR #291, opened 2021-06-02
Fix password reading in AutoYaST

imobachgs:
* Do not pretend to read the /etc/shadow line anymore.
* Make a difference between nil and empty values.
* See https://github.com/yast/yast-users/pull/285#pullrequestreview-669800708 for context.

### DIFF
--- a/src/lib/y2users/autoinst/reader.rb
+++ b/src/lib/y2users/autoinst/reader.rb
@@ -79,37 +79,25 @@ module Y2Users
         end
       end
 
-      # shadow attrs without password and username which is done manually
-      SORTED_SHADOW_ATTRS = [
-        # Looks like shadow last change is not part of User.Export TODO: verify
-        "last_change", "min", "max", "warn", "inact", "expire", "flag"
-      ].freeze
-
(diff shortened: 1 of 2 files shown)


src/lib/y2users/autoinst/reader.rb
```diff
@@ -79,37 +79,25 @@ def read_groups
...
-      def shadow_string(user)
-        other_attrs = if user.password_settings
-          SORTED_SHADOW_ATTRS.map { |a| user.password_settings.send(a) }
-        else
-          []
-        end
-
-        [user.username, user.user_password, *other_attrs].join(":")
+        password.aging = PasswordAging.new(password_settings.last_change)
+        password.minimum_age = password_settings.min
```
joseivanlopez, line 95: Do those methods `password_settings.min`, etc return a string?

imobachgs: A string or nil, if it is not even present. Is it OK?

joseivanlopez: Perfect! I was wondering in case they return an integer.

PR #293, opened 2021-06-02
Y2users linuxrc root pwd

jreidinger:
trello: https://trello.com/c/CT6tRt83/2484-1-autoyast-read-root-password-from-booting-command-line

tested beside unit tests also manually

### DIFF
--- a/src/lib/users/clients/auto.rb
+++ b/src/lib/users/clients/auto.rb
@@ -131,7 +131,11 @@ module Y2Users
           config.attach(root_user)
         end
 
-        root_user.password = Y2Users::Password.create_plain(Yast::Linuxrc.InstallInf("RootPassword"))
+        root_user.password = Y2Users::Password.create_plain(
+          Yast::Linuxrc.InstallInf("RootPassword")
+        )
+
+        root_user
       end
     end
   end
--- a/src/lib/users/clients/auto.rb
+++ b/src/lib/users/clients/auto.rb
@@ -128,7 +128,7 @@ module Y2Users
         if !root_user
           root_user = Y2Users::User.new("root")
           root_user.uid = "0"
-          config.users.add(root_user)
+          config.attach(root_user)
         end
 
         root_user.password = Y2Users::Password.create_plain(Yast::Linuxrc.InstallInf("RootPassword"))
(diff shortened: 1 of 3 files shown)


src/lib/users/clients/auto.rb
```diff
@@ -116,6 +118,25 @@ def check_users(users)
 
         Yast::Report.Error(_("Found users in profile with equal <uid>.")) if report
       end
+
+      def read_linuxrc_root_pwd(config)
+        root_user = config.users.root
+        # use param only if profile do not already contain it
```
joseivanlopez, line 124: Wrong english: only if profile does not contain it yet.

src/lib/users/clients/auto.rb
```diff
@@ -116,6 +118,25 @@ def check_users(users)
...
+        return if root_user&.password
+
+        # root user not defined
+        if !root_user
```
joseivanlopez, line 128: This is not the first time we need to populate a config with a root user, see https://github.com/yast/yast-users/blob/y2users/src/lib/y2users/users_simple/reader.rb#L71.

Maybe we need a method to generate the default root user. Otherwise, we can end up with root users populated with different data. What do you think?

jreidinger: in fact we do not need default root user but just set root user password. so maybe helper in UsersCollection or Config makes sense? But I am not so sure.

joseivanlopez: Maybe something like `config.attach(User.create_root)` ?

jreidinger: well, that already imply we need to check existence of root and do some actions. Do we ever use `#root` as nil? maybe UsersCollection.root can add root if it is not there yet?

joseivanlopez: I would keep the collections agnostic to the business logic. I would delegate the responsibility of having a root or not to the user of the class (as it is now). The only thing I would like to have is a root template for the cases where we decide to add a root to the collection. Anyway, I think this detail can be addressed later, I don't want to block the PR only for that ;).

joseivanlopez (approved):
LGTM

PR #294, opened 2021-06-02
Remove dependency of `Y2Users::HelpTexts` on `UsersSimple`

dgdavid:
### Problem

`Y2Users::HelpTexts` depends on `Yast::UsersSimple` just for the `Yast::UsersSimple.ValidPasswordHelptext`, which simply returns a localized string.

https://trello.com/c/j0t4rJAZ/2430-1-remove-dependency-of-y2usershelptexts-on-userssimple

### Solution

As we agreed, for this special case the simpler solution is just to duplicate the localized string in `Y2Users::HelpText`.

### Tests

Tested manually checking that the translation still working as expected.

passwords_help_text

### DIFF
--- a/src/lib/y2users/help_texts.rb
+++ b/src/lib/y2users/help_texts.rb
@@ -26,7 +26,6 @@ module Y2Users
   # moving stuff to the Y2Users namespace
   module HelpTexts
     include Yast::I18n
-    Yast.import "UsersSimple"
 
     # Needed to be able to call textdomain below
     extend Yast::I18n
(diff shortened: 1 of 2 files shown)


ancorgs (approved):
LGTM. But maybe we should add a comment in both `#valid_password_text` and `#ValidPPasswordHelpText` mentioning the existence of each other.

src/lib/y2users/help_texts.rb
```diff
@@ -47,7 +46,15 @@ def validation_config
     #
     # @return [String] formatted and localized text
     def valid_password_text
-      Yast::UsersSimple.ValidPasswordHelptext
+      # NOTE: to avoid dependencies, this help text is a duplicate of
+      #   Yast::UsersSimple.ValidPPasswordHelpText. Please, keep them in sync.
```
ancorgs, line 50: ValidPPassword, double "P"

dgdavid: :see_no_evil: I'll hide your comment right away! It never happened :stuck_out_tongue_winking_eye:

ancorgs (approved):
LGTM

jreidinger (approved):

PR #296, opened 2021-06-03
[AutoYaST] Report issues when importing the users/groups list

imobachgs:
Trello: https://trello.com/c/uSZUbp5E/
See https://github.com/yast/yast-autoinstallation/pull/766

* Instead of just failing at the end of the installation, report as soon as possible missing values for `groupname` or `username` using the `Y2Issues` reporting mechanism.
* Adds a `ReadResult` class that encapsulates the configuration and the list of detected issues. At this point, only `Autoinst::Reader#read` returns an instance of this class.
* Although the auto client returns `false` when the user does not want to continue, we need to fix AutoYaST to not ignore this value.

## Example

Given the following profile:

```xml
  <users config:type="list">
    <user>
      <encrypted config:type="boolean">false</encrypted>
      <user_password>linux</user_password>
      <username>root</username>
    </user>
    <user t="map">
      <fullname>SUSE</fullname>
      <user_password>linux</user_password>
      <encrypted config:type="boolean">false</encrypted>
    </user>
  </users>

  <groups config:type="list">
    <group t="map">
      <gid>101</gid>
    </group>
  </groups>
```
It causes the following error:

users-issues


src/lib/y2users/autoinst/reader.rb
```diff
@@ -99,6 +105,41 @@ def create_password(user)
         password.account_expiration = AccountExpiration.new(password_settings.expire)
         password
       end
+
+      # Validates the user and adds the problems found to the given issues list
+      #
+      # @param user_section [Y2Users::AutoinstProfile::UserSection] User section from the profile
+      # @param issues [Y2Issues::List] Add issues list
```
dgdavid, line 112: NP: the word "Add" looks wrong there.

src/lib/y2users/autoinst/reader.rb
```diff
@@ -99,6 +105,41 @@ def create_password(user)
...
+        return true if user_section.username && !user_section.username.empty?
+
+        issues << invalid_value_issue(user_section, :username)
+        false
+      end
+
+      # Validates the group and adds the problems found to the given issues list
+      #
+      # @param group_section [Y2Users::AutoinstProfile::GroupSection] User section from the profile
+      # @param issues [Y2Issues::List] Add issues list
```
dgdavid, line 124: NP: Same than above

src/lib/y2users/read_result.rb
```diff
@@ -0,0 +1,44 @@
...
+
+module Y2Users
+  # Class to represent the result of reading the configuration
+  class ReadResult
+    attr_reader :config, :issues
+
+    # Represents a reading operation result
+    #
+    # @param config [Config] Read configuration
+    # @param issues [Errors::List] Errors list
```
dgdavid, line 31: Did you mean?

```suggestion
    # @param issues [Y2Issues::List] Errors list
```

ancorgs:
Shouldn't this be somehow integrated with the existing AutoinstIssues mechanism?

For example, if both the storage section (which uses AutoinstIssues) and the users section contain errors, the user will get two separate pop-ups, instead of a summary of all the Autoinstallation issues. Is that by design?

src/lib/y2users/autoinst/reader.rb
```diff
@@ -99,6 +105,41 @@ def create_password(user)
         password.account_expiration = AccountExpiration.new(password_settings.expire)
         password
       end
+
+      # Validates the user and adds the problems found to the given issues list
+      #
+      # @param user_section [Y2Users::AutoinstProfile::UserSection] User section from the profile
+      # @param issues [Y2Issues::List] Add issues list
+      # @return [Boolean] true if the user is valid; false otherwise.
+      def valid_user?(user_section, issues)
```
ancorgs, line 114: NP: Not sure if I like these boolean methods called `valid_bla?` that, in addition to returning true or false, populate the given issues list. But I guess that's something we will fix with https://trello.com/c/yVrdlyCx/4747-refactor-linuxwriter (I added a paragraph about having the readers also into account).

imobachgs: Well, I am not a big fan, but it is a private API so I am fine with it. Additionally, a user can be valid although it contains issues.

imobachgs:
> Shouldn't this be somehow integrated with the existing AutoinstIssues mechanism?
> 
> For example, if both the storage section (which uses AutoinstIssues) and the users section contain errors, the user will get two separate pop-ups, instead of a summary of all the Autoinstallation issues. Is that by design?

Yes, it is how it works by now. It is something to improve in the future, but not now.

src/lib/y2users/read_result.rb
```diff
@@ -0,0 +1,44 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "y2issues"
+
+module Y2Users
+  # Class to represent the result of reading the configuration
+  class ReadResult
```
joseivanlopez, line 25: NP: the equivalent class in *network* is called [`ReadingResult`](https://github.com/yast/yast-network/blob/master/src/lib/y2network/reading_result.rb). Maybe we should use the same name here, what do you think?

imobachgs: Well, to be honest, I prefer `ReadResult`, so I guess we should adapt network instead. :-)

dgdavid (approved):
Since https://github.com/yast/yast-users/pull/296#issuecomment-854452896 will be addressed in future PR, this still LGTM.

@ancorgs, @joseivanlopez, Do you agree?

ancorgs (approved):
LGTM, assuming the unification with AutoinstIssues will come soon(ish).

PR #297, opened 2021-06-08
Y2users ruby validation

jreidinger:
trello: https://trello.com/c/yDXqaGHa/2429-2-ruby-based-validators


ancorgs:
I guess we are just duplicating the strings here to reuse the translations. Have you tried if the translations are indeed reused?

Regarding that, we maybe should do like in #294 and comment the duplication

test/lib/y2users/user_validator_test.rb
```diff
@@ -75,6 +75,15 @@
         end
       end
 
+      context "and using invalid characters" do
+        # See UserValidator::CHAR_REGEXP
+        let(:username) { "abc++" }
+
+        it "includes a fatal issue for the username length" do
```
dgdavid, line 82: Description looks wrong. What about using something like 


```suggestion
        it "includes a fatal issue for using not allowed characters" do
```

src/lib/y2users/user_validator.rb
```diff
@@ -90,5 +91,127 @@ def add_fatal_issue(list, error, location)
...
+    LETTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz".freeze
+    NUMBERS = "0123456789".freeze
+    # Regexp for allowed characters.
+    # NOTE: this is based on default in login.defs, maybe read it on running system?
+    CHAR_REGEXP = /\A[#{LETTERS}_][#{LETTERS}#{NUMBERS}_.-]*[#{LETTERS}#{NUMBERS}_.$-]?\z/.freeze
+    def check_characters
+      return "" if user.name =~ CHAR_REGEXP
+
+      _("The username may contain only\n" \
+        "letters, digits, \"-\", \".\", and \"_\"\n" \
```
mvidner, line 116: In Russian, Я is a letter.

```suggestion
        "Latin letters and digits, \"-\", \".\", and \"_\"\n" \
```

src/lib/y2users/user_validator.rb
```diff
@@ -90,5 +91,127 @@ def add_fatal_issue(list, error, location)
...
+
+    LETTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz".freeze
+    NUMBERS = "0123456789".freeze
+    # Regexp for allowed characters.
+    # NOTE: this is based on default in login.defs, maybe read it on running system?
+    CHAR_REGEXP = /\A[#{LETTERS}_][#{LETTERS}#{NUMBERS}_.-]*[#{LETTERS}#{NUMBERS}_.$-]?\z/.freeze
+    def check_characters
+      return "" if user.name =~ CHAR_REGEXP
+
+      _("The username may contain only\n" \
```
mvidner, line 115:
```suggestion
      # TRANSLATORS: take care to preserve the ASCII characters
      #  - (U+002D HYPHEN-MINUS)  . (U+002E FULL STOP) _ (U+005F UNDERSCORE),
      #  not replacing them with your native script variants
      _("The username may contain only\n" \
```

src/lib/y2users/user_validator.rb
```diff
@@ -90,5 +91,127 @@ def add_fatal_issue(list, error, location)
...
+        "username and an existing username.\n" \
+        "Try another one.")
+    end
+
+    def check_full_name
+      return "" unless user.full_name
+
+      return "" if user.full_name !~ /[:,]/
+
+      _("The user's full name cannot contain\n" \
```
mvidner, line 212:
```suggestion
      # TRANSLATORS: take care to preserve the ASCII characters
      #  : (U+003A COLON)  , (U+002C COMMA),
      #  not replacing them with your native script variants
      _("The user's full name cannot contain\n" \
```

src/lib/y2users/password_validator.rb
```diff
@@ -68,6 +69,20 @@ def issues
...
+    LETTERS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ".freeze
+    NUMBERS = "0123456789".freeze
+    # TODO: find source for this info beside bsc#175706
+    PASSWORD_REGEXP = /\A[-#{NUMBERS}#{LETTERS}!@#\:!\$%^&*() ,;:._+\/|?{}=\['\"`~<>\]\\]+\z/.freeze
+    def check_password
+      return _("No password entered.\nTry again.") if !content || content.empty?
+
+      return "" if content =~ PASSWORD_REGEXP
+
+      _("The password may only contain the following characters:\n" \
```
mvidner, line 81: This message needs to be adapted according to what we find about the non-allowed characters.
Something like

> The allowed characters are ... ... ...
> (which are the printable ASCII characters except ... and ...)

src/lib/y2users/user_validator.rb
```diff
@@ -47,21 +48,21 @@ def issues(skip: [])
...
       if !skip.include?(:name)
-        err = Yast::UsersSimple.CheckUsernameLength(user.name)
+        err = check_length
         add_fatal_issue(list, err, NAME_LOC)
 
-        err = Yast::UsersSimple.CheckUsernameContents(user.name, "")
+        err = check_characters
         add_fatal_issue(list, err, NAME_LOC)
 
         # Yast::UsersSimple.CheckUsernameConflicts is currently used only when manually creating
```
mvidner, line 57: Is the comment still valid?

mvidner (changes requested):

src/lib/y2users/password_validator.rb
```diff
@@ -68,6 +69,20 @@ def issues
     # @return [Y2Users::User] user containing the password to validate
     attr_reader :user
 
+    LETTERS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ".freeze
+    NUMBERS = "0123456789".freeze
+    # TODO: find source for this info beside bsc#175706
+    PASSWORD_REGEXP = /\A[-#{NUMBERS}#{LETTERS}!@#\:!\$%^&*() ,;:._+\/|?{}=\['\"`~<>\]\\]+\z/.freeze
```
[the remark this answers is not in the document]
jreidinger: to be honest I have no clue, I just duplicate what is working in UsersSimple

src/lib/y2users/user_validator.rb
jreidinger: good note...so we need to change it in both UsersSimple and here.

src/lib/y2users/password_validator.rb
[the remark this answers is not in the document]
mvidner: Good point, login.defs has a comment saying why, which we should repeat here:
In non-English locales, A-Z includes different letters than the ASCII range. Notably, in Estonian it excludes T, because Z is a "foreign letter" shoved right after S in the alphabet.

src/lib/y2users/user_validator.rb
jreidinger: yes, it still applies, as I just duplicate userssimple functionality in context of y2users, so installation and autoinstallation for local users. So it is more like note when we want to check duplicities in running system that it has same limitation that checks against hard coded list.

src/lib/y2users/password_validator.rb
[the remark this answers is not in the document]
jreidinger: due to locales. As e.g. in czech a-z include also ščřčžřáýžááíý characters

test/lib/y2users/user_validator_test.rb
jreidinger: copy and paste error, thanks for catch

src/lib/y2users/password_validator.rb
```diff
@@ -68,6 +69,20 @@ def issues
...
+    def check_password
+      return _("No password entered.\nTry again.") if !content || content.empty?
```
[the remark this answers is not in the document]
jreidinger: I really look forward when ruby contain rails blank? method :)

jreidinger:
@ancorgs good question, I checked it for users validation and it worked. But now when I check also password validation, there is a difference probably caused by some escaping, will fix it and also add comments.

src/lib/y2users/password_validator.rb
jreidinger: One topic I would like to discuss on call is this unallowed characters as I cannot find reason for it beside not so easy to type so special ones. So maybe it should not be fatal and just warn user that password contain special or non-ascii characters that can be hard to type during login?

src/lib/y2users/password_validator.rb
```diff
@@ -68,6 +69,23 @@ def issues
...
 
+    LETTERS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ".freeze
+    NUMBERS = "0123456789".freeze
+    # TODO: find source for this info beside bsc#175706
+    PASSWORD_REGEXP = /\A[-#{NUMBERS}#{LETTERS}!@#\:!\$%^&*() ,;:._+\/|?{}=\['\"`~<>\]\\]+\z/.freeze
+    def check_password
+      # note duplicite string as in UsersSimple. Keep them in sync.
+      return _("No password entered.\nTry again.") if !content || content.empty?
+
+      return "" if content =~ PASSWORD_REGEXP
```
joseivanlopez, line 80: NP: I would return nil when there is no error.

src/lib/y2users/password_validator.rb
mvidner: Yes, basically this is to avoid a situation where the user puts non-ASCII characters in the password and then cannot log in at the console because for whatever reason the keyboard layout is wrong / defaults to en_US.
We actually did get such a bug report from a German user.

src/modules/UsersSimple.pm
```diff
@@ -411,9 +411,11 @@ my $valid_logname_chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQ
 
 my $valid_password_chars = "[-0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!@#\$%^&*() ,;:._+/|?{}=\['\"`~<>]|]";# the ']' is or-ed...
 
-# error popup	
+# TRANSLATORS: Keep symbols as it is as it is ASCII variants of symbols, not UTF-8 ones.
```
joseivanlopez, line 414: Typo: as it is as it is

jreidinger: well, in fact it is not typo. first "as it is" I use to keep it intact and second "as it is" is explanation why to keep it intact, but I rephrase it as it can be confusing and I am sure native English speaker won't use it :wink:

src/lib/y2users/password_validator.rb
```diff
@@ -68,6 +69,26 @@ def issues
...
+
+      # note duplicite string as in UsersSimple. Keep them in sync. Be aware of different
+      # escaping in perl
+      _("The password may only contain the following characters:\n" \
+        "0-9, a-z, A-Z, and any of \"`~!@#$%^&* ,.;:._-+/|?='{[(<>)]}\\\".\n" \
+        "Basically it is available keys on default keyboard layout to prevent\n" \
```
mvidner, line 87:
```suggestion
        "Basically these are the available keys on the default keyboard layout to prevent\n" \
```

src/lib/y2users/password_validator.rb
```diff
@@ -68,6 +69,26 @@ def issues
...
+        "problems when emergency login from console is needed.\n" \
```
mvidner, line 88:
```suggestion
        "problems when an emergency login from the console is needed.\n" \
```

src/modules/UsersSimple.pm
```diff
@@ -411,9 +411,11 @@ my $valid_logname_chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQ
 
 my $valid_password_chars = "[-0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!@#\$%^&*() ,;:._+/|?{}=\['\"`~<>]|]";# the ']' is or-ed...
 
-# error popup	
+# TRANSLATORS: Keep symbols in its ASCII variants that can be preset on default us keyboard layout
 my $valid_password_message = __("The password may only contain the following characters:
 0-9, a-z, A-Z, and any of \"`~!\@#\$%^&* ,.;:._-+/|\?='{[(<>)]}\\\".
+Basically it is available keys on default keyboard layout to prevent
```
mvidner, line 417:
```suggestion
Basically these are the available keys on the default keyboard layout to prevent
```

src/modules/UsersSimple.pm
```diff
@@ -411,9 +411,11 @@ my $valid_logname_chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQ
...
+problems when emergency login from console is needed.
```
mvidner, line 418:
```suggestion
problems when an emergency login from the console is needed.
```

mvidner (changes requested):
Thanks, just some grammared the adding.

mvidner (approved):

src/lib/y2users/user_validator.rb
```diff
@@ -47,21 +48,21 @@ def issues(skip: [])
...
         # the initial user during installation, it simply checks against a hard-coded list of
         # system user names that are expected to exist in a system right after installation.
-        err = Yast::UsersSimple.CheckUsernameConflicts(user.name)
+        err = check_username_conflict
```
ancorgs, line 60: As pointed in the comment, "Yast::UsersSimple.CheckUsernameConflicts is currently used only when manually creating the initial user". So we can remove the Perl function now.

jreidinger: thanks for notice, done in https://github.com/yast/yast-users/pull/308

marcusmueller:
Because the question came up [here](https://github.com/yast/yast-users/pull/297), why was the check for usernames to be at least two characters long introduced? I assume that is discussed behind the seemingly dead trello link, and thus not documented publicly anymore.

jreidinger:
@marcusmueller I would say it was it for very very long time...so probably for some really historical reason and we just keep it in new ruby implementation

jreidinger:
but it is true, that I do not find reason to have 2 as minimal size.
The latest commit I can find regarding having minimal length `2` is 20 years old without reasoning https://github.com/yast/yast-users/blame/SLE-10-SP1/src/UsersCache.pm#L73

PR #295, opened 2021-06-03
Process user_defaults section from the AutoYaST profile

ancorgs:
Still WIP.

For details, see https://trello.com/c/QU2gDrTK/2486-3-autoyast-userdefaults-section

### Included/Planned

* [x] Previous partial polishing of the Linux Writer
* [x] Add `Y2Users::User#skel`
* [ ] `Autoinst::Reader` configure the users based on "skel", "groups" and "no_groups"
* [ ] `Autoinst::Reader` generate a `Config#useradd` based on the profile content
* [ ] Make sure target_config.useradd makes after merging it with ay_config

PR #298, opened 2021-06-09
Process user_defaults section from the AutoYaST profile

ancorgs:
This replaces #295.

For details, see https://trello.com/c/QU2gDrTK/2486-3-autoyast-userdefaults-section

Better review commit by commit (see checklist below)

## Status

This was closed because manual testing proved that the list of secondary groups configured via "groups" and "no_groups" has never been honored by AutoYaST when creating the initial users. That list is ignored in SLE-11-SPX, SLE-12-SPX and SLE-15-SPX, all cases verified with extensive manual tests.

So there is no point in incorporating code to support something we can remove and nobody will miss. Quite the opposite, it's not unlikely that if we fix AutoYaST to honor the secondary groups (and, thus, the AutoYaST documentation) we would end up with users complaining about the behavior change.

### Included

* [x] Previous partial polishing of the Linux Writer
* [x] Add `Y2Users::UserDefaults` as a superset of `Y2Users::UseraddConfig`
* [x] `Autoinst::Reader` generates a full-featured `Config#user_defaults` based on the profile content
* [x] Make sure target_config.user_defaults makes after merging it with ay_config
* [x] `Linux::Writer` honors all the attributes in `Config#user_defaults` (not only the useradd ones)

### Testing

- Unit tests included
- Manual testing WIP


src/lib/y2users/autoinst/reader.rb
```diff
@@ -81,6 +87,11 @@ def read
       # @return [AutoinstProfile::LoginSettingsSections]
       attr_reader :login_settings_section
 
+      # Profile section describing the default configuration for new users
+      #
+      # @return [AutoinstProfile::UserDefaultsSections]
```
joseivanlopez, line 92: Typo: UserDefaultsSection~~s~~

src/lib/y2users/config.rb
```diff
@@ -145,8 +145,20 @@ def merge!(other)
 
     # Useradd configuration to be applied to the system before creating the users
     #
-    # @return [UseraddConfig, nil] nil if the configuration is unknown
-    attr_accessor :useradd
+    # @see #user_defaults
+    #
+    # @return [UserDefaults, nil] nil if the configuration is unknown
```
joseivanlopez, line 150: The return value should be a `UseraddConfig`.

test/lib/y2users/autoinst/reader_test.rb
```diff
@@ -49,6 +49,14 @@
       expect(root_user.password.account_expiration.expire?).to eq(false)
 
       expect(config.login?).to eq(false)
+
+      user_defaults = config.user_defaults
+      expect(user_defaults.expiration).to eq nil
+      expect(user_defaults.inactivity_period).to eq(-1)
```
joseivanlopez, line 55: This kind of values are strings in `Password` class. Could it be a problem?

joseivanlopez (approved):
LGTM. Minor suggestions and a question.

src/lib/y2users/linux/login_config_writer.rb
```diff
@@ -0,0 +1,58 @@
...
+
+module Y2Users
+  module Linux
+    # Configures the login settings in the target system according to a given {LoginConfig} object
+    class LoginConfigWriter
+      Yast.import "Autologin"
+
+      # Constructor
+      #
+      # @param login [LoginConfig] see #login_config
```
jreidinger, line 30: I think below is mentioned also nil as possible value when nothing is to write. So just add it to possible types.

src/lib/y2users/linux/users_writer.rb
```diff
@@ -0,0 +1,366 @@
...
+
+      # Executes the command for creating the user
+      #
+      # @param user [User] the user to be created on the system
+      # @param issues [Y2Issues::List] a collection for adding an issue if something goes wrong
+      def add_user(user, issues)
+        Yast::Execute.on_target!(USERADD, *useradd_options(user))
+        change_password(user, issues) if user.password
+        write_auth_keys(user, issues)
+      rescue Cheetah::ExecutionFailed => e
```
jreidinger, line 119: I have to say that this desing kind of confuse me as I thought that it catched execute failure also for chpasswd and others. I think it kind of hit by issue that it is different level of abstraction. Maybe remove this rescue completely and separate Yast::Execute to own method with own rescue? And here have just
```ruby
call_useradd(user, issues)
change_password(user, issues) if user.password
write_auth_keys(user, issues)
```

and when we are at it I see also that if inconsistent and probably should also live in change_password method.

src/lib/y2users/linux/users_writer.rb
```diff
@@ -0,0 +1,366 @@
...
+      # @param old_user [User] Original user
+      # @param issues [Y2Issues::List] a collection for adding an issue if something goes wrong
+      def edit_user(new_user, old_user, issues)
+        usermod_changes = USERMOD_ATTRS.any? do |attr|
+          !new_user.public_send(attr).nil? &&
+            (new_user.public_send(attr) != old_user.public_send(attr))
+        end
+        usermod_changes ||= different_groups?(new_user, old_user)
+
+        Yast::Execute.on_target!(USERMOD, *usermod_options(new_user, old_user)) if usermod_changes
```
jreidinger, line 205: and same what I mention above also here. Here it is even more visible as it contain also usermod detection.

src/lib/y2users/linux/users_writer.rb
```diff
@@ -0,0 +1,366 @@
...
+        opts << user.name
+        opts
+      end
+
+      # Name of the secondary groups for useradd
+      #
+      # @return [Array<String>]
+      def useradd_groups(user)
+        names = user.secondary_groups_name
+        return names unless names.empty?
```
jreidinger, line 245: Here I see a bit issue. Can we distinguish between explicit no groups and intentionally skipping of definition of secondary groups?

src/lib/y2users/linux/users_writer.rb
```diff
@@ -0,0 +1,366 @@
...
+      end
+
+      # Command to modify the user
+      #
+      # @param new_user [User] User containing the updated information
+      # @param old_user [User] Original user
+      # @return [Array<String>] usermod options
+      # rubocop:disable Metrics/CyclomaticComplexity
+      # rubocop:disable Metrics/AbcSize
+      # rubocop:disable Metrics/PerceivedComplexity
```
jreidinger, line 257: so much disable raise a bit my internal alarm if we can do it better :)

I see basically two options to reduce its complexity:

1. has method like `modified?(new_user, old_user, :gid)` that can be used in that if ( only exception is gecos which we talk we should change to nil approach if not defined and different_groups? method
2. define map that defines options and conditions. Something like
```ruby
USERMOD_OPTIONS = [
  Proc.new { |o, n| (o.gid != n.gid && n.gid) ?  ["--gid", n.gid] : [] },
  ...
  Proc.new { |o, n| different_groups?(new_user, old_user) ? ["--groups", new_user.secondary_groups_name.join(",")] : []
]
```

But to be honest I do not like neither of it. But I think it missing research did by @dgdavid that said that passing same values are harmless as usermod itself detects it. So maybe condition can be simplified to check for nil.

src/lib/y2users/linux/users_writer.rb
```diff
@@ -0,0 +1,366 @@
...
+      end
+
+      # Options for `useradd` to create the home directory
+      #
+      # @param _user [User]
+      # @return [Array<String>]
+      def create_home_options(_user)
+        # TODO: "--btrfs-subvolume-home" if needed
+        opts = ["--create-home"]
+        opts.concat(["--skel", defaults.skel]) if defaults.forced_skel?
```
jreidinger, line 301: just style matter but it is inconsistent with rest of file that use `<<` approach for adding options.

src/lib/y2users/linux/users_writer.rb
```diff
@@ -0,0 +1,366 @@
...
+
+      # Generates and returns the options expected by `chpasswd` for the given user
+      #
+      # @param user [User]
+      # @return [Array<String, Hash>]
+      def chpasswd_options(user)
+        opts = []
+        opts << "-e" if user.password&.value&.encrypted?
+        opts << {
+          stdin:    [user.name, user.password&.value&.content].join(":"),
```
jreidinger, line 313: I see that `&.` usage as kind of strange, as I am not sure if chpasswd survives. ( my test run failed due to pam settings, but maybe it can survive to set passwordless login which is maybe even more dangerous from security POV?

test/lib/y2users/autoinst/reader_test.rb
ancorgs: I don't think so, but I will double-check

src/lib/y2users/linux/login_config_writer.rb
ancorgs: I will. But, again, let's try to not re-review code that was just moved from the original `Linux::Writer` to other files. We may loose the focus.

src/lib/y2users/linux/users_writer.rb
ancorgs: Again, this is not new code. Just moved to another file. Let's stay focused on this PR by not re-reviewing that. What about a separate card for this kind of things we want to revisit?

src/lib/y2users/linux/users_writer.rb
ancorgs: This is again a re-review of old code that was just moved. Let's keep that out for the time being.

src/lib/y2users/linux/users_writer.rb
ancorgs: Again, old code. Just copied as-is (including comments... and rubocop comments). Let's take care in a separate card/PR, please.

src/lib/y2users/linux/users_writer.rb
ancorgs: Again, this is old code that was just moved.

src/lib/y2users/linux/users_writer.rb
ancorgs: I will change it

src/lib/y2users/autoinst/reader.rb
ancorgs: Correct. I just realized I made the typo by copy-pasting from the line above (LoginSettingsSections)

src/lib/y2users/linux/users_writer.rb
jreidinger: I am not sure if I get your comment that `There is no way in AutoYaST to specify the list of secondary groups for a user (not even to force an empty one).` In https://doc.opensuse.org/projects/autoyast/#Configuration-Security-groups there is attribute userlist that specify secondary groups for given user. Or do I miss something? But I agree that it is hard to distinguish in autoyat itself if there is intentionally no secondary groups or if default should be used :worried:

ancorgs: Fun fact. I did some manual testing with the current AutoYaST (before any y2users change) and... ta da! the default list of groups is always ignored.

- If the user is listed in the attribute "userlist" of some group(s), then the internal attribute "grouplist" for that user is initialized according. As a result, the code does not enter [here](https://github.com/yast/yast-users/blob/master/src/modules/Users.pm#L3119). That means the default group list is not used.
- If the user is not listed in any group, the internal attribute "grouplist" is initialized to an empty set. Which means it's defined and the code does not enter the if either. So again the default group list is not used.

We can conclude the "groups" attribute of the `<user_defaults>` section of AutoYaST has been useless since useradd dropped the GROUPS key from `/etc/defaults/useradd`. So the only use case for which we have been implementing all this has disappeared. :disappointed:

ancorgs: > We can conclude the "groups" attribute of the `<user_defaults>` section of AutoYaST has been useless since useradd dropped the GROUPS key from `/etc/defaults/useradd`.

In case you wonder. That means with the switch from SLE11 to SLE12 (as documented somewhere in this PR).

src/lib/y2users/linux/users_writer.rb
dgdavid: Just for the record, I changed my mind few days after doing such research. I mean, I prefer to provide only modified attributes to the command, no matter what it does currently.

ancorgs:
Closed in favor of #306 after finding out the support for secondary groups in AutoYaST has been broken since (at least) SLE-11-SP2.

PR #299, opened 2021-06-09
Drop UsersSimple

joseivanlopez:
## Problem

During the installation, the information about the users to create/import is stored in the *UsersSimple* perl module and recovered later in order to write it in the system. This is done by means of a specific reader/writer. But *UsersSimple* module is not needed at all because the information about users could be stored in the `ConfigManager`.

## Solution

Remove usage of `UsersSimple` module during the installation in favor of `ConfigManager`.

Some notes:

* The installation is not perl-free yet, because some features provided by *UsersSimple* (e.g., loading cracklib) are still needed.
* In the running system, *UsersSimple* is still used for validations.

Other modules depending of *UsersSimple* have been adapted too. These PRs should be merged after merging this changes into master:

* https://github.com/yast/yast-firewall/pull/150
* https://github.com/yast/yast-installation/pull/969
* https://github.com/yast/yast-firstboot/pull/123

## Tests

* Added unit tests
* Manually tested
* Manually tested using AutoYaST


src/include/users/dialogs.rb
```diff
@@ -1120,16 +1121,9 @@ def EditUserDialog(what)
...
-              full = UsersSimple.Transliterate(full)
-              UI.ChangeWidget(
-                Id(:username),
-                :Value,
-                Builtins.tolower(
-                  Builtins.filterchars(full, UsersSimple.ValidLognameChars)
-                )
-              )
+              full_name = UI.QueryWidget(Id(ret), :Value)
+              username = Y2Users::UsernameGenerator.for(full_name)
```
joseivanlopez, line 1125: Thinking twice, maybe a class is not needed at all. A module would be enough. And in case of using a module, I see no problem with naming it as `Username`. Something like:

~~~
module Username
  def generate_from(value)
    transliterate(value)....
  end
end
~~~

And then here:

~~~
username = Yast2::Username.generate_from(full_name)
~~~

What do you think?

src/lib/users/proposal.rb
```diff
@@ -29,6 +32,9 @@ def initialize
       Yast.import "Wizard"
       Yast.import "Mode"
       textdomain "users"
+
+      @users_event_id = "users--user"
+      @root_event_id = "users--root"
```
jreidinger, line 37: why not constant?

src/lib/users/proposal.rb
```diff
@@ -78,18 +81,17 @@ def ask_user(param)
...
+      menu = []
 
       if !Mode.auto
-        menu = [ # menu button label
-          { "id" => "users--user", "title" => _("&User") },
+        id = users_event_id
+        menu = [
           # menu button label
-          { "id" => "users--root", "title" => _("&Root Password") },
+          { "id" => "users--user", "title" => _("&User") },
```
jreidinger, line 91: using constant here?

src/lib/users/proposal.rb
```diff
@@ -102,69 +104,171 @@ def description
...
 
+    attr_reader :users_event_id, :root_event_id
+
+    # The config holding users and groups to create
+    #
+    # @return [Y2Users::Config]
+    def config
+      return @config if @config
+
+      # FIXME: should AutoYaST use `target` configuration as well?
```
jreidinger, line 115: ideally yes and at the end always merge target and system based on strategy for given mode.

src/lib/users/proposal.rb
```diff
@@ -102,69 +104,171 @@ def description
...
-      format(_("Password Encryption Method: <a href=%s>%s</a>"),
-        "users--encryption", ::Users::EncryptionMethod.current.label)
+    # Text to display when users are being imported
+    #
+    # @see #users_proposal
+    # @return [String]
+    def imported_users_summary
+      # TRANSLATORS: summary line,
+      #   %{hs} and %{he} are the hyperlink start and end respectively.
+      #   %{qty} will be replacrd by the number of users
```
jreidinger, line 235: typo replacrd

src/lib/users/proposal.rb
joseivanlopez: Adapted

jreidinger (approved):

PR #300, opened 2021-06-10
add config manager target helper

jreidinger:


src/lib/y2users/config_manager.rb
```diff
@@ -81,5 +81,19 @@ def system(reader: nil, force_read: false)
...
+    def target
+      config(:target)
+    end
+
+    # Sets target modification used to store configuration before it is applied
+    # @param [Config] config Note that config should not be frozen as it is expected to be modified
+    # @note in various scenarious target does not need to be exact target state, but just
+    # interstate that is then merged with real system state
+    def target=(config)
+      register(config, as: :target)
```
dgdavid, line 96: Should we check that it is a Y2Users::Config before registering it?

dgdavid (approved):
It LGTM as far as we are going to discard the idea of starting #target [with some content](https://github.com/yast/yast-users/blob/497449c3815d2bfb833e972c7b4b98edeff2cdba/src/lib/y2users/config_manager.rb#L86-L95) if it does not exist yet.

Just have a look to my question about checking the given config param and, if you don't mind, wait for @joseivanlopez comments since he said "will comment later" in IRC.

Thanks!

src/lib/y2users/config_manager.rb
jreidinger: well, we do not do it also in register...if we do it in register, we get it for free here. So to be consistent just do what register do.

dgdavid: Right!

src/lib/y2users/config_manager.rb
```diff
@@ -81,5 +81,19 @@ def system(reader: nil, force_read: false)
 
       config
     end
+
+    # Gets target modification used to store configuration before it is applied
```
joseivanlopez, line 85: NP: target modification? I don't understand it very well. I would rephrase it a bit: Config used to store the configuration to apply.

src/lib/y2users/config_manager.rb
```diff
@@ -81,5 +81,19 @@ def system(reader: nil, force_read: false)
...
+    # @return [Config,nil] returns nil if no target is yet initialized
+    def target
+      config(:target)
+    end
+
+    # Sets target modification used to store configuration before it is applied
```
joseivanlopez, line 91: NP: what about: Sets the target configuration, see {#target}.

src/lib/y2users/config_manager.rb
```diff
@@ -81,5 +81,19 @@ def system(reader: nil, force_read: false)
...
+    # @param [Config] config Note that config should not be frozen as it is expected to be modified
+    # @note in various scenarious target does not need to be exact target state, but just
```
joseivanlopez, line 93: NP: I would move this note to the documentation of `#target`.

joseivanlopez (approved):
LGTM. Only comments about the documentation.

PR #301, opened 2021-06-10
Remove some dead code

dgdavid:
[`users/routines`](https://github.com/yast/yast-users/blob/8d0a665f1423a4439f143b957e3c96b114c6f6ae/src/include/users/routines.rb#L109-L176) contains two methods that are no longer needed after https://github.com/yast/yast-users/commit/1b6818ce05463a28471171686f0f6a57c69107e3 and https://github.com/yast/yast-users/commit/4a5ec436df0e191c4e13bf43290c793d03008f60. So, let's drop them.

---

Note: the only [client using `#setup_all_users`](https://github.com/yast/yast-firstboot/blob/b550444a77ffc410aec0e130f8ca0e9fdf445c45/src/lib/y2firstboot/clients/user.rb#L44) lives in yast2-firsboot and will be adapted.

### DIFF
--- a/src/include/users/routines.rb
+++ b/src/include/users/routines.rb
@@ -137,42 +137,4 @@ module Yast
       end
     end
   end
-
-  # setup ALL users (included root user, autologin, root aliases,...)
-  # Return: true if there has been added a user
-  def setup_all_users
-    ret = false
-
-    # disable UI (progress)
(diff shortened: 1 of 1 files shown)


jreidinger (approved):

joseivanlopez (approved):
LGTM

PR #307, opened 2021-06-15
Y2users auto write

jreidinger:
trello: https://trello.com/c/Vwj139MR/2434-3-autoyast-in-installed-system-write-users-using-y2users

In the end after several loops the change is mainly about keeping previous "working" state of autoyast UI.

Tested manually.


src/lib/users/clients/auto.rb
```diff
@@ -46,17 +46,23 @@ def initialize
...
+        if Yast::Stage.initial
+          reader = Y2Users::Autoinst::Reader.new(param)
+          result = reader.read
+          read_linuxrc_root_pwd(result.config)
+
+          if result.issues?
+            return false unless Y2Issues.report(result.issues) == :yes
+          end
+
+          Y2Users::ConfigManager.instance.register(result.config, as: :autoinst)
```
joseivanlopez, line 59: Not related to this PR, but maybe we should start using `target` for the autoyast config too.

src/lib/users/clients/auto.rb
```diff
@@ -105,7 +112,7 @@ def modified
       end
 
       def reset
-        Y2Users::ConfigManager.instance.unregister(:autoinst)
+        import({})
```
imobachgs, line 115: np: the result might be different, although I do not think it is a problem. When calling reset, you will have an empty configuration at :autoinst, or am I missing something? Is it relevant? I do not think so, but perhaps it is not the expected result of calling `#reset`.

joseivanlopez (approved):
LGTM

src/lib/users/clients/auto.rb
jreidinger: this is basically revert to previous version as reset is called only in autoyast UI. So I keep it as broken as it was before.

jreidinger: so `:autoinst` case should not happen as it is `config` mode.

src/lib/users/clients/auto.rb
jreidinger: yep, but as said on call my goal is to not break autoyast UI, so I mainly copy existing stuff without touching it.

src/lib/users/clients/auto.rb
imobachgs: OK, fine with me. I was simply wondering if it was correct from a logic point of view.

PR #308, opened 2021-06-15
drop no longer used perl check method

jreidinger:
forgotten drop. see https://github.com/yast/yast-users/pull/297#discussion_r651884383


joseivanlopez:
We are cleaning up the UsersSimple module as part of https://github.com/yast/yast-users/pull/299, so this method will be dropped anyway.

jreidinger:
@joseivanlopez ok, if you also adapt comment respectively I can close this PR.

PR #309, opened 2021-06-16
Small fixes 

ancorgs:
Fixing a bug found during manual testing of the AutoYaST installation... and adding the debug information that leaded me to finding the problem.

### DIFF
--- a/src/lib/y2users/group.rb
+++ b/src/lib/y2users/group.rb
@@ -82,7 +82,16 @@ module Y2Users
     def users
       return [] unless attached?
 
-      config.users.select { |u| u.gid == gid || users_name.include?(u.name) }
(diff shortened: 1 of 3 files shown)


imobachgs (approved):

jreidinger (approved):

src/lib/y2users/group.rb
```diff
@@ -82,7 +82,16 @@ def initialize(name)
     def users
       return [] unless attached?
 
-      config.users.select { |u| u.gid == gid || users_name.include?(u.name) }
+      config.users.select { |u| same_gid?(u) || users_name.include?(u.name) }
+    end
+
+  private
+
+    # @see #users
```
jreidinger, line 90: np: I would maybe just explain why two nils are not same gid ( group nil is not yet assigned and user nil means do not change )

PR #311, opened 2021-06-17
Ids collisions

jreidinger:
trello: https://trello.com/c/tRAkWt82/2503-3-dealing-with-possible-duplicity-in-uids-guids

screenshot of final testing result with my testing profile:

- there are two users collisions
- there are one group collision
- nscd failed because I try change home to its symlink location and useradd failed to create it ( so nothing bad )
- root user failed because due to modified home directory that is in use in chroot ( known issue in y2users that old implementation does not have, not sure how big issue it is )

ids_collision

After notice from @imobachgs I found that it does not use predefined presenter, so I try to modify code to use it and UI looks better for me now:

ids_collision2


src/lib/users/clients/users_finish.rb
```diff
@@ -93,6 +95,12 @@ def target_config
       @target_config
     end
 
+    def check_ids
+      validator = Y2Users::IdsValidator.new(Y2Users::ConfigManager.instance.target)
```
joseivanlopez, line 99: I think we should use `target_config` (which is already merged with system config) instead of `Y2Users::ConfigManager.instance.target`. Otherwise, it will not detect ids duplicity with existing users/groups.

jreidinger: I think it is job of merger to avoid such duplicitites. That is reason to use user config  and not result of merger. But if you think it will be better to report it to user in advance also about system collision, I can pass that.

joseivanlopez: The merger merges users/groups by name, but it checks nothing about ids. I think it would be nice to notice to the user in advance if there is an id duplicity with the existing users/groups.

joseivanlopez: Ok, I got the point, the autoyast merger reuses the id.

dgdavid:
I have tested it _manually_ using a provided profile via IRC. Collisions were properly detected, see picture below

autoyast-collisions

However, please note that some unit tests still failing https://github.com/yast/yast-users/pull/311/checks?check_run_id=2913720286#step:5:1553

Could you please have a look?

src/lib/y2users/collision_checker.rb
```diff
@@ -0,0 +1,127 @@
...
+
+      list
+    end
+
+  private
+
+    # @return [Y2Users::Config] config to validate
+    attr_reader :config
+
+    def duplicite_users_ids
```
ancorgs, line 65: duplicite? Is that even an English word?

src/lib/y2users/collision_checker.rb
```diff
@@ -0,0 +1,127 @@
...
+require "y2issues"
+
+module Y2Users
+  # Internal class to check if there are collisions in config.
+  # This is not part of the stable API.
+  class CollisionChecker
+    include Yast::I18n
+
+    # Issue location describing the Group#gid attribute
+    GROUP_ID_LOC = "group:gid".freeze
```
ancorgs, line 29: I know this `CollisionChecker` is designed to be generic. But at the end of the day, it's detecting a problem in the AutoYaST profile, so I would expect the location to point to the problematic part of the profile. That is, stuff like `autoyast:users` or `autoyast:groups`.

src/lib/y2users/collision_checker.rb
```diff
@@ -0,0 +1,127 @@
...
+        msg = format(_("Groups %{groups} have same GID %{gid}."),
+          groups: groups.map(&:name).join(", "),
+          gid:    gid)
+        Y2Issues::Issue.new(msg, location: GROUP_ID_LOC, severity: :warn)
+      end
+
+      Y2Issues::List.new(issues)
+    end
+
+    def duplicite_groups_names
```
ancorgs, line 111: NP: don't you feel the code of the four methods is too similar? Cannot the logic be extracted to some base method?

src/lib/y2users/collision_checker.rb
ancorgs: According to the screenshot provided by @dgdavid, I'm not the only one who cannot make sense of the current locations, the Y2Issues presenter seems to be struggling with them as well.

src/lib/y2users/collision_checker.rb
jreidinger: french one, I will fix it.

src/lib/y2users/collision_checker.rb
jreidinger: well, I think that it can also report issues on running system, but if it helps autoyast to group it as autoyast:users it is ok for me.

ancorgs:
This is how it looks with the latest changes:

ids

dgdavid:
@ancorgs, I see it a bit different:

Screenshot_openSUSE-Leap_2021-06-25_14:53:04

dgdavid:
See how it looks after https://github.com/yast/yast-users/pull/311/commits/cdff1baf7f3e728d361c3787fdf2c1e5f446ec1f

collision-issues-no-location

ancorgs (approved):
LGTM

PR #313, opened 2021-06-23
Add y2users documentation

imobachgs:
* Import documentation from the yast-users-ng repository.


joseivanlopez (approved):
LGTM

doc/use-cases-old-module.md
```diff
@@ -0,0 +1,351 @@
+# Use-cases of the Old YaST Users Module
```
ancorgs, line 1: This mentions the old yast module assuming a new one is to be created. That's not the case anymore.

ancorgs (approved):
LGTM

PR #316, opened 2021-06-25
Merge the Y2Users branch into master

ancorgs:
YaST does not rely on `useradd` or similar tools to create, modify and delete users and groups. Instead, it performs by itself all the changes in the system like creating the home directories, assigning uids/gids or modifying the files /etc/passwd, /etc/groups, etc.

But the behavior of `useradd` has changed over time:

- It was totally rewritten in the jump from SLE-11 to SLE-12 (in which it was also moved from the package `pwdutils` to the current `shadow` one).
- It changed the way the skeleton directories are created.
- It added management of subids and subgids.
- It changed the location of its configuration, with some attributes removed from `/etc/defaults/useradd`, either moved to a different file or simply gone.

None of those changes were added to YaST over the years, so the behavior of both tools has diverged. That has caused many problems in the past (see links below) and could cause more in the future. So the YaST team decided to rewrite the management of users to rely more on `useradd` and its associated tools in the `shadow` package.

All that was developed in a separate `y2users` branch on this repository. This pull request merges all those changes into master.

### Relevant issues/feature requests

* [PM-2620](https://jira.suse.com/browse/PM-2620), [bsc#1166743](https://bugzilla.suse.com/show_bug.cgi?id=1166743) and [bsc#1185342](https://bugzilla.suse.com/show_bug.cgi?id=1185342) - Different management of subuids between YaST and useradd
* [bsc#1183136](https://bugzilla.suse.com/show_bug.cgi?id=1183136) and [bsc#1179261](https://bugzilla.suse.com/show_bug.cgi?id=1179261) - Different management of skel between YaST and useradd
* [bsc#1099153](https://bugzilla.suse.com/show_bug.cgi?id=1099153) - UMASK moved to another file

## Included in this branch / pull request

- YaST now relies on `useradd`, `groupadd` and other `shadow` tools when creating users and groups during installation and also when using AutoYaST or Firstboot.
- YaST uses the new Y2Issues mechanism to report errors to the user.
- Implemented a new object-oriented API to read and manage local users (Y2Users).
- Adapted the management of the `<user_defaults>` section of the AutoYaST profile to recent versions of useradd (implies dropping support for `<groups>`, `<no_groups>` and `<skel>`). See https://github.com/SUSE/doc-sle/pull/901 for the rationale and historic background of each change.


dgdavid (approved):
Let's go!

mvidner:
Yayy!

This card summarizes the remaining dependencies on Perl code, in case we want to stop depending on Perl: https://trello.com/c/hTxvdapk

PR #317, opened 2021-07-14
System settings warning

joseivanlopez:
## Problem

System settings are read when the *users* client starts. But the client is not aware of failures when reading the system settings. The client silently continues, and it applies its own default values for the system settings (e.g., for login.defs values). Everything happens without noticing to the user.

Moreover, the default encryption method is md5 when *login.defs* cannot be read. This is insecure, and sha512 should be the default encryption method.

* https://bugzilla.suse.com/show_bug.cgi?id=1177183

## Solution

A warning popup is shown when reading system settings fails. And sha512 is the default encryption method now. 

* Requires https://github.com/yast/yast-security/pull/111

## Testing

* Manually tested

## Screenshots

Screenshot from 2021-07-14 15-31-06

Screenshot from 2021-07-14 15-31-30

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jul 14 14:18:32 UTC 2021 - José Iván López González <jlopez@suse.com>
+
+- Show warning when reading system settings (e.g., login.defs)
+  fails (bsc#1177183).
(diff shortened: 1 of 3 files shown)


imobachgs (approved):

joseivanlopez:
https://build.suse.de/request/show/245217

PR #318, opened 2021-07-15
Port PR#317 to master

joseivanlopez:
See https://github.com/yast/yast-users/pull/317.

### DIFF
--- a/package/yast2-users.spec
+++ b/package/yast2-users.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-users
-Version:        4.4.2
+Version:        4.4.3
 Release:        0
 Summary:        YaST2 - User and Group Configuration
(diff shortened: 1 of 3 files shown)


package/yast2-users.spec
```diff
@@ -49,7 +49,8 @@ Requires:       yast2-country
 # CFA::Nsswitch
 Requires:       yast2-pam >= 4.3.0
 
-Requires:       yast2-security
+# Security::SafeRead
+Requires:       yast2-security >= 4.3.17
```
imobachgs, line 53: You should use version `4.4.1`, because `4.4.0 > 4.3.17` but it does not work.

imobachgs (changes requested):

imobachgs (approved):

PR #319, opened 2021-07-16
Improve knowing when first user is being created or edited

dgdavid:
### Problem

Reported [bsc#1188068](https://bugzilla.suse.com/show_bug.cgi?id=1188068#c12) has revealed that `InstUserFirstDialog` is using a not optimal way to know if the user is going to be created or edited (see comment#12). In fact, changes introduced at https://github.com/yast/yast-firstboot/pull/126 broken it in a way that [`init_pw_for_root`](https://github.com/yast/yast-users/blob/61c0bc6bf2bb595f2d86fe37807868194c44d865/src/lib/users/dialogs/inst_user_first.rb#L195-L209) starts failing.

### Solution

To use the [`Y2Users::ConfigElement#attached?`](https://github.com/yast/yast-users/blob/61c0bc6bf2bb595f2d86fe37807868194c44d865/src/lib/y2users/config_element.rb#L65) method instead of relying on the password value, which looks prone to false positives.

However, the check has to be performed and stored at the time to initialize the dialog because the private [InstUserFirsDialogt#user](https://github.com/yast/yast-users/blob/61c0bc6bf2bb595f2d86fe37807868194c44d865/src/lib/users/dialogs/inst_user_first.rb#L211-L220) method attaches the user to the current configuration for validation purposes.

### Tests

* Already unit tests still passing
* Manual tests: done via driver update in a TW installation. Everything seems work as expected.

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jul 16 13:13:26 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- Use a more reliable way of knowing if editing a user (related to
+  bsc#1188068).
+- 4.4.4
+
+-------------------------------------------------------------------
 Wed Jul 14 14:18:32 UTC 2021 - José Iván López González <jlopez@suse.com>
 
 - Show warning when reading system settings (e.g., login.defs)
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,8 +1,8 @@
 -------------------------------------------------------------------
 Fri Jul 16 13:13:26 UTC 2021 - David Diaz <dgonzalez@suse.com>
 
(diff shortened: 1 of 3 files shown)


joseivanlopez:
Looks good.

package/yast2-users.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Jul 16 13:13:26 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- Use a more reliable way of knowing if editing a user (related to
```
joseivanlopez, line 4: NP: I would reword it a bit: "Use a more reliable way of knowing if a user is being edited".

joseivanlopez (approved):
LGTM

package/yast2-users.changes
dgdavid: Ok, done.

PR #320, opened 2021-07-19
Improve handling of authorized_keys

dgdavid:
### Problem

As [stated in the documentation](https://documentation.suse.com/sles/15-SP3/html/SLES-all/cha-configuration-installation-options.html#id-1.7.5.2.35.3.5), the `$HOME/.ssh/authorized_keys` file will be overwritten when auto-installing the product via an AutoYaST profile defining a set of authorized_keys

> If the profile defines a set of SSH authorized keys for a user in the authorized_keys section, an existing $HOME/.ssh/authorized_keys file will be overwritten. If not existing, the file will be created with the content specified. Avoid overwriting an existing authorized_keys file by not specifying the respective section in the AutoYaST control file. 

However, as reported at https://bugzilla.suse.com/show_bug.cgi?id=1188361, looks like the file is overwritten even when the profile does not define authorized keys

### Solution

* [x] Do not overwrite the file when none authorized keys has been defined

### Tests

* [x] Unit tests
* [x] Manual testing when (auto)installing with a profile defining authorized_keys for a user already having a `$HOME/.ssh/authorized_keys` file
* [x] Manual testing when (auto)installing with a profile not defining authorized_keys for a user already having a `$HOME/.ssh/authorized_keys` file
* [x] Manual test in a running system to check that everything work as expected while editing the SSH authorized keys for a user.

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jul 20 15:09:28 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- Do not rewrite authorized_keys unless it is needed (bsc#1188361).
+- 4.2.13
+
+-------------------------------------------------------------------
 Tue Jun 15 09:17:24 UTC 2021 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Fix the Comment entry in the desktop file so the tooltip
--- a/src/lib/users/ssh_authorized_keyring.rb
+++ b/src/lib/users/ssh_authorized_keyring.rb
@@ -218,13 +218,8 @@ module Yast
       def write_file(owner, group)
         file = SSHAuthorizedKeysFile.new(authorized_keys_path)
         file.keys = keys
-
         log.info "Writing #{keys.size} keys in #{authorized_keys_path}"
-
-        return unless file.save
-
-        FileUtils::Chown("#{owner}:#{group}", authorized_keys_path, false)
-        FileUtils::Chmod(AUTHORIZED_KEYS_PERMS, authorized_keys_path, false)
+        file.save && FileUtils::Chown("#{owner}:#{group}", authorized_keys_path, false)
       rescue SSHAuthorizedKeysFile::NotRegularFile
         raise NotRegularAuthorizedKeysFile, authorized_keys_path
       end
--- a/src/lib/users/ssh_authorized_keyring.rb
+++ b/src/lib/users/ssh_authorized_keyring.rb
@@ -218,8 +218,13 @@ module Yast
       def write_file(owner, group)
         file = SSHAuthorizedKeysFile.new(authorized_keys_path)
         file.keys = keys
+
         log.info "Writing #{keys.size} keys in #{authorized_keys_path}"
-        file.save && FileUtils::Chown("#{owner}:#{group}", authorized_keys_path, false)
+
+        return unless file.save
+
+        FileUtils::Chown("#{owner}:#{group}", authorized_keys_path, false)
+        FileUtils::Chmod(AUTHORIZED_KEYS_PERMS, authorized_keys_path, false)
       rescue SSHAuthorizedKeysFile::NotRegularFile
         raise NotRegularAuthorizedKeysFile, authorized_keys_path
       end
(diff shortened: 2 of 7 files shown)


test/lib/users/ssh_authorized_keyring_test.rb
```diff
@@ -90,122 +90,179 @@ def authorized_keys_from_home(path)
...
-        expect(Yast::FileUtils).to receive(:Chown)
-          .with("#{uid}:#{gid}", authorized_keys_path, false)
+      context "but new keys are added" do
+        before do
+          keyring.add_keys([key])
+        end
 
-        keyring.write_keys
-      end
+        it "deletes old authorized_keys file" do
```
imobachgs, line 194: np: I know it is something from the past, but whether the file is removed or not (before recreating it) is basically an implementation detail.

src/lib/users/ssh_authorized_keyring.rb
```diff
@@ -218,8 +218,13 @@ def create_ssh_dir(user, group)
...
         file = SSHAuthorizedKeysFile.new(authorized_keys_path)
         file.keys = keys
+
         log.info "Writing #{keys.size} keys in #{authorized_keys_path}"
-        file.save && FileUtils::Chown("#{owner}:#{group}", authorized_keys_path, false)
+
+        return unless file.save
+
+        FileUtils::Chown("#{owner}:#{group}", authorized_keys_path, false)
+        FileUtils::Chmod(AUTHORIZED_KEYS_PERMS, authorized_keys_path, false)
```
imobachgs, line 227: Just a note (we already discussed it offline): is not needed at all: https://github.com/yast/yast-users/blob/7735c840cc98327f526a731e5ff87769d906b79b/src/lib/users/ssh_authorized_keys_file.rb#L119. Actually, the `AUTHORIZED_KEYS_PERMS` looks like a left-over.

imobachgs (changes requested):
In general, it looks good. Just a comment about file permissions.

package/yast2-users.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Jul 20 15:09:28 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- Improve handling of authorized_keys file for recreating it
```
imobachgs: I would say "Do not rewrite authorized_keys unless it is needed.". Or even "Do not modify authorized_keys...".

src/lib/users/ssh_authorized_keyring.rb
```diff
@@ -150,8 +163,6 @@ def write_keys
       AUTHORIZED_KEYS_FILE = "authorized_keys".freeze
       # @return [String] Permissions to be set on SSH_DIR directory
       SSH_DIR_PERMS = "0700".freeze
-      # @return [String] Permissions to be set on `authorized_keys` file
-      AUTHORIZED_KEYS_PERMS = "0600".freeze
```
imobachgs, line 154: :+1:

imobachgs (approved):
IMHO, as soon as we confirm it is working in our manual tests, it can be merged.

dgdavid:
@imobachgs, 

I'm asking for a review again because of https://github.com/yast/yast-users/pull/320/commits/5a9985f30348ae727b8e0a25b840998440311239, needed to have the intended changes working during installation. Please, have a look.

dgdavid:
> To be honest, it is quite unexpected for me that the `SSHAuthorizedKeys#write_keys` read the keys. Does it mean that it will always keep the keys that are already there?

That's exactly what I was testing just after write the comment (but it took me some time and I have to postpone tests until now, sorry). I can confirm now that https://github.com/yast/yast-users/commit/5a9985f30348ae727b8e0a25b840998440311239  breaks the behavior in the running system and also the old testsuite.

So let's discuss it tomorrow morning, if you don't mind.

src/lib/users/ssh_authorized_keyring.rb
```diff
@@ -115,27 +116,39 @@ def empty?
...
 
       # Write user keys to the given file
       #
       # If SSH_DIR does not exist in the given directory, it will be
       # created inheriting owner/group and setting permissions to SSH_DIR_PERM.
       def write_keys
+        return unless changed?
+
         remove_authorized_keys_file
+
```
imobachgs, line 144: np: why this blank line?

imobachgs (approved):

src/lib/users/ssh_authorized_keyring.rb
dgdavid: Nothing special. I just think that it helps to read the code better. I'll keep it, hope you don't mind :)

PR #323, opened 2021-08-12
Extend Linux writer

joseivanlopez:
### Problem

Right now, the `Linux::Writer` class only supports (auto)installation and firstboot use cases. During the installation, users can only be created. And for firstboot and auto-installation, users can be edited too. In firstboot, even the name of the user can be changed when you go back, but the `Linux::Writer` does not modify it in the system.

* https://bugzilla.suse.com/show_bug.cgi?id=1188612

Moreover, the root password is not properly deleted if you go back to the wizard step for setting the root password and remove password (only an authorized key is selected). This affects to both: installation and firstboot.

* https://bugzilla.suse.com/show_bug.cgi?id=1189402

### Solution

Extend `Linux::Writer` in order to support the use cases mentioning above.

See also https://github.com/yast/yast-firstboot/pull/127.

### Testing

* Added unit tests.
* Manually tested.

### DIFF
--- a/src/lib/users/widgets.rb
+++ b/src/lib/users/widgets.rb
@@ -102,11 +102,18 @@ module Users
       Yast::UI.ChangeWidget(Id(:pw2), :Value, current_password)
     end
 
+    # FIXME: Validation methods should only check values. Note that with the current implementation,
+    #   this method is taking care of setting/removing the user's password. This method should not
+    #   modify the user.
     # rubocop:disable Metrics/CyclomaticComplexity
     def validate
       password1 = Yast::UI.QueryWidget(Id(:pw1), :Value)
       password2 = Yast::UI.QueryWidget(Id(:pw2), :Value)
-      return true if allow_empty? && password1.empty?
+
+      if allow_empty? && password1.empty?
+        @user.password = nil
+        return true
+      end
 
       if password1 != password2
         # report misspellings of the password
--- a/src/lib/y2users/linux/users_writer.rb
+++ b/src/lib/y2users/linux/users_writer.rb
@@ -105,6 +105,10 @@ module Y2Users
       CHPASSWD = "/usr/sbin/chpasswd".freeze
       private_constant :CHPASSWD
 
+      # Command for editing a password (i.e., used for deleting the password)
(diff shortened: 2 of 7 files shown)


src/lib/users/widgets.rb
```diff
@@ -102,11 +102,18 @@ def init
       Yast::UI.ChangeWidget(Id(:pw2), :Value, current_password)
     end
 
+    # FIXME: Validation methods should only check values. Note that with the current implementation,
+    #   this method is taking care of setting/removing the user's password. This method should not
+    #   modify the user.
```
dgdavid, line 107: :+1: :+1: :+1:

src/lib/y2users/linux/users_writer.rb
```diff
@@ -241,6 +260,17 @@ def edit_user(new_user, old_user, issues)
         log.error("Error modifying user '#{new_user.name}' - #{e.message}")
       end
 
+      # Edits the user's password
+      #
+      # @param new_user [User] User containing the updated information
+      # @param old_user [User] Original user
+      # @param issues [Y2Issues::List] a collection for adding an issue if something goes wrong
+      def edit_password(new_user, old_user, issues)
+        return if old_user.password == new_user.password
```
shundhammer, line 269: Is it possible that the admin might want to change only the _other_ fields in `/etc/passwd` with this (real name etc.), but the password remains the same? In this case, those other changes would simply be ignored if the password is not also changed.

dgdavid (approved):
LGTM, thanks a lot.

src/lib/y2users/linux/users_writer.rb
joseivanlopez: The password is an object, and the equality is overloaded, see https://github.com/yast/yast-users/blob/master/src/lib/y2users/password.rb#L76. So, the passwords are not equal if any of the other fields is different.

dgdavid: I don't think so. This guard clause applies only to the `edit_password` method. Or am I missing something?

shundhammer: OK, thanks for the clarification.

PR #326, opened 2021-09-15
Linux writer improvements

joseivanlopez:


src/lib/y2users/linux/helpers.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2issues"
+
+module Y2Users
+  module Linux
+    # Helper methods for readers and writers
+    module Helpers
```
jreidinger, line 25: To be honest I do not like much any "Helpers" as it tends to grow a lot over time. Why not having simple mixing like "WithIssues" that adds the method for issues handling? I think it can be even useful generally in yast2 for other modules.

joseivanlopez: Good point.

jreidinger (approved):

PR #327, opened 2021-09-16
Document supported attributes

joseivanlopez:
Add document to describe the correspondence between `Yast::Users` data structure and `Y2Users` attributes.

### DIFF
--- a/doc/attributes.md
+++ b/doc/attributes.md
@@ -1,0 +1,66 @@
+# User And Group Attributes
+
+This document describes the attributes that can be indicated for users and groups, either through the YaST UI of the *users* client or through AutoYaST. Each attribute should have a counterpart in the new *Y2Users* code. This correspondence helps to translate the users data structure from *Yast::Users* perl module to the new code.
+
+## User
+
+The YaST UI for creating and editing a user offers the following attributes:
+
+| UI Field | Y2Users attr | Only when adding | Only when editing | Description |
+| :--- | :--- | :--- | :--- | :--- |
+| full name | gecos | | | |
+| name | name | | | |
+| password | password | | | |
+| system mail | | | | |
+| disable login | * password starting by `!` | | | |
+| uid | uid | | | |
+| home dir | home | | | | |
+| home dir permission | | | | |
+| empty home | | yes | | |
+| Move to new location | | | yes |  |
+| Btrfs subvolume | btrfs_subvolume_home | | | |
+| Additional info | gecos | | | |
+| login shell | shell | | | |
+| group | primary_group | | | |
+| additional groups | groups | | | |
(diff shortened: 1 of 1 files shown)


doc/attributes.md
```diff
@@ -0,0 +1,51 @@
...
+
+The YaST UI for creating and editing a user offers the following attributes:
+
+| UI Field | Y2Users attr | Only when adding | Only when editing | Description |
+| :--- | :--- | :--- | :--- | :--- |
+| full name | gecos | | | |
+| name | name | | | |
+| password | password | | | |
+| system mail | | | | |
+| disable login | * password starting by !$ | | | |
```
dgdavid, line 15: More exactly, password starting with `!`. See [usermod(8) man page](https://linux.die.net/man/8/usermod)

> ...
> 
> **-L, --lock**
    Lock a user's password. This puts a '!' in front of the encrypted password, effectively disabling the password. You can't use this option with -p or -U.
> 
>     Note: if you wish to lock the account (not only access with a password), you should also set the EXPIRE_DATE to 1. 
> 
> ...
> 
> **-U, --unlock**
    Unlock a user's password. This removes the '!' in front of the encrypted password. You can't use this option with -p or -L.
> 
>    Note: if you wish to unlock the account (not only access with a password), you should also set the EXPIRE_DATE (for example to 99999, or to the EXPIRE value from /etc/default/useradd).

doc/attributes.md
```diff
@@ -0,0 +1,51 @@
...
+| login shell | shell | | | |
+| group | primary_group | | | |
+| additional groups | groups | | | |
+
+### Home management
+
+### Creating a user
+
+* A user can be created with or without home.
+* To create a user without home, the *home dir* path should be set to empty.
```
dgdavid, line 32: That's true... and tricky.

Out of the scope of this document, of course, but I'd fix the UI to make it explicit. Right now, if user clean the _home dir_ but goes back to the _User Data_ tab and forward to _Details_ tab... the *home dir* is auto-filling again.

doc/attributes.md
```diff
@@ -0,0 +1,51 @@
...
+* The home can be created with or without default content (*empty home* checkbox).
+* The home can be created as Btrfs subvolume (*Btrfs subvol* checkbox).
+
+### Editing a user
+
+* If the home path is changed, then a new home is created with the default content.
+* There is no way to create a new home without content (*empty home* checkbox is not available).
+* The old home is removed.
```
dgdavid, line 40: From the top of my head, I believe that not always, Only when moving the content to a new location

dgdavid (approved):
I just added some comments just in case they are useful, but document itself looks good to me.

Thanks a lot for taking care of writing it. I own you a beer :) :)

PR #328, opened 2021-09-16
Add reader from Users module

jreidinger:
- [x] tests
- [x] manual testing with changes in runtime UI
- [x] remove testing reader run

### DIFF
--- a/src/lib/y2users/users_module/reader.rb
+++ b/src/lib/y2users/users_module/reader.rb
@@ -123,6 +123,9 @@ module Y2Users
         # set password only if specified, nil means not touch it
         user.password = create_password(user_attrs) if user_attrs["userPassword"]
 
+        # set authorized keys only if set in users module
+        user.authorized_keys = user_attrs["authorized_keys"] if user_attrs["authorized_keys"]
+
         user
       end
 
--- a/src/lib/y2users/users_module/reader.rb
+++ b/src/lib/y2users/users_module/reader.rb
@@ -108,7 +108,7 @@ module Y2Users
         user
       end
 
-      # Creates a {Group} object based on the data structure of a Users user
+      # Creates a {Group} object based on the data structure of a Users group
       #
       # @param group_attrs [Hash] a group representation in the format used by Users
       # @return [Group]
--- a/src/lib/y2users/users_module/reader.rb
+++ b/src/lib/y2users/users_module/reader.rb
@@ -26,6 +26,7 @@ require "y2users/login_config"
 require "y2users/useradd_config"
 
 Yast.import "Users"
+Yast.import "Autologin"
 
 module Y2Users
   module UsersModule
@@ -37,7 +38,7 @@ module Y2Users
       def read
         Config.new.tap do |config|
           read_elements(config)
-          # TODO: read auto-login, but it is not exposed
+          config.login = read_autologin
           config.useradd = read_useradd_defaults
         end
       end
@@ -62,6 +63,23 @@ module Y2Users
         UseraddConfig.new(y2users_map)
       end
 
+      # Reads the Autlogin config from Autologin module
+      #
+      # @note Users reads and sets autologin directly and do not expose it
+      #   so it is easier to read it directly
+      #
+      # @return [LoginConfig]
+      def read_autologin
+        res = LoginConfig.new
+        res.autologin_user = if Yast::Autologin.user.empty?
+            nil
+          else
+            Yast::Autologin.user
+          end
+        res.passwordless = Yast::Autologin.pw_less
+
+        res
+      end
 
       # Reads the users and groups
       #
@@ -169,5 +187,6 @@ if $0 == __FILE__
   require "pp"
 
   Yast::Users.Read
+  Yast::Autologin.Read
   pp Y2Users::UsersModule::Reader.new.read
 end
--- a/src/lib/y2users/users_module/reader.rb
+++ b/src/lib/y2users/users_module/reader.rb
@@ -1,0 +1,173 @@
+# Copyright (c) [2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "y2users/config"
+require "y2users/user"
+require "y2users/group"
+require "y2users/password"
+require "y2users/login_config"
+require "y2users/useradd_config"
+
+Yast.import "Users"
+
+module Y2Users
+  module UsersModule
+    # Class for reading the configuration from Yast::Users module
+    class Reader
+      # Generates a new config with the information from YaST::Users module
+      #
+      # @return [Config]
+      def read
+        Config.new.tap do |config|
+          read_elements(config)
+          # TODO: read auto-login, but it is not exposed
+          config.useradd = read_useradd_defaults
+        end
+      end
+
+    private
+
+      # Reads the useradd defaults from Users module
+      #
+      # @return [UseraddConfig]
+      def read_useradd_defaults
+        users_map = Yast::Users.GetLoginDefaults
+        # notes: groups are deprecated and usrskell and create_mail_spool is not in users
+        y2users_map = {
+          group: users_map["group"],
+          home: users_map["home"],
+          umask: users_map["umask"],
+          expiration: users_map["expire"],
+          inactivity_period: users_map["inactive"],
+          shell: users_map["shell"],
+          skel: users_map["skel"],
+        }
+        UseraddConfig.new(y2users_map)
+      end
+
+
+      # Reads the users and groups
+      #
+      # @param config [Config]
+      def read_elements(config)
+        config.attach(users)
+        config.attach(groups)
+      end
+
+      # Returns the list of users
+      #
+      # @return [Array<User>] the collection of users
+      def users
+        all_local = Yast::Users.GetUsers("uid", "local").values +
+          Yast::Users.GetUsers("uid", "system").values
+        all_local.map { |u| user(u) }
+      end
+
+      def groups
+        all_local = Yast::Users.GetGroups("cn", "local").values +
+          Yast::Users.GetGroups("cn", "system").values
+        all_local.map { |u| group(u) }
+      end
+
+      # Creates a {User} object based on the data structure of a Users user
+      #
+      # @param user_attrs [Hash] a user representation in the format used by Users
+      # @return [User]
+      def user(user_attrs)
+        user = User.new(user_attrs["uid"])
+
+        user.uid   = attr_value(:uidNumber, user_attrs)
+        user.gid   = attr_value(:gidNumber, user_attrs)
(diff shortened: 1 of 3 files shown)


src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,173 @@
...
+          user.system = true
+        end
+
+        # set password only if specified, nil means not touch it
+        user.password = create_password(user_attrs) if user_attrs["userPassword"]
+
+        user
+      end
+
+      # Creates a {Group} object based on the data structure of a Users user
```
dgdavid, line 111:
```suggestion
      # Creates a {Group} object based on the data structure of a Users group
```

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,173 @@
...
+        # TODO: group passwords not supported
+        # group.password = create_password(group_attrs) if group_attrs["userPassword"]
+
+        group
+      end
+
+      # Value of the given field for the given Users user
+      #
+      # @param attr [#to_s] name of the attribute
+      # @param user [Hash] a user representation in the format used by Users
```
dgdavid, line 136: A user or group, I guess

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,173 @@
...
+    end
+  end
+end
+
+if $0 == __FILE__
+  require "pp"
+
+  Yast::Users.Read
+  pp Y2Users::UsersModule::Reader.new.read
+end
```
dgdavid, line 173: NP: to be removed :wink:

dgdavid (commented):
I have to play with it a little bit more to get more familiar. But initially, despite those small neat-pickings from my side, it starts looking good.

src/lib/y2users/users_module/reader.rb
dgdavid: Ah, it's in your check list in the description

> remove testing reader run

I overlooked it

jreidinger: yep, it is in todo. But very useful for quick testing

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,187 @@
...
+      end
+
+    private
+
+      # Reads the useradd defaults from Users module
+      #
+      # @return [UseraddConfig]
+      def read_useradd_defaults
+        users_map = Yast::Users.GetLoginDefaults
+        # notes: groups are deprecated and usrskell and create_mail_spool is not in users
```
dgdavid, line 53:
```suggestion
        # notes: groups are deprecated and usrkel and create_mail_spool is not in users
```

NOTE: skel and create_mail_spool are, however, read from the `useradd` configuration file. See https://github.com/yast/yast-users/blob/29a8e9402e76b25313c159c671866e9d61eb7f1a/src/lib/y2users/useradd_config.rb#L154-L179

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,187 @@
...
+          umask:             users_map["umask"],
+          expiration:        users_map["expire"],
+          inactivity_period: users_map["inactive"]&.to_i,
+          shell:             users_map["shell"],
+          skel:              users_map["skel"]
+        }
+        UseraddConfig.new(y2users_map)
+      end
+
+      # Reads the Autlogin config from Autologin module
```
dgdavid, line 66:
```suggestion
      # Reads the Autologin config from Autologin module
```

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,187 @@
...
+        user.uid   = attr_value(:uidNumber, user_attrs)&.to_s
+        user.gid   = attr_value(:gidNumber, user_attrs)&.to_s
+        user.gid ||= config.groups.by_name(attr_value["groupname"])&.gid
+
+        user.shell = attr_value(:loginShell, user_attrs)
+        user.home  = attr_value(:homeDirectory, user_attrs)
+        user.gecos = [attr_value(:cn, user_attrs), *user_attrs["addit_data"].split(",")].compact
+        user.system = true if !user.uid && user_attrs["type"] == "system"
+
+        # set password only if specified, nil means not touch it
```
dgdavid, line 123: Aren't we are just reading current user values at this point? I mean, setting (i.e., changing) the password depending on if we have a value or `nil` is something for the writer, isn't it?

dgdavid (commented):
Just a few minor comments fro my side

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,187 @@
...
+
+      # Reads the Autlogin config from Autologin module
+      #
+      # @note Users reads and sets autologin directly and do not expose it
+      #   so it is easier to read it directly
+      #
+      # @return [LoginConfig]
+      def read_autologin
+        res = LoginConfig.new
+        res.autologin_user = if Yast::Autologin.user.empty?
```
joseivanlopez, line 74: NP: why not in one line? `res.autologin_user = Yast::Autologin.user.empy? ? nil : Yast::Autologin.user`

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,187 @@
...
+  module UsersModule
+    # Class for reading the configuration from Yast::Users module
+    class Reader
+      # Generates a new config with the information from YaST::Users module
+      #
+      # @return [Config]
+      def read
+        Config.new.tap do |config|
+          read_elements(config)
+          config.login = read_autologin
```
joseivanlopez, line 41: NP: due to `config` is passed to `#read_elements`, maybe I would pass it to `#read_autologin` and `read_useradd_defaults` too.

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,187 @@
...
+          inactivity_period: users_map["inactive"]&.to_i,
+          shell:             users_map["shell"],
+          skel:              users_map["skel"]
+        }
+        UseraddConfig.new(y2users_map)
+      end
+
+      # Reads the Autlogin config from Autologin module
+      #
+      # @note Users reads and sets autologin directly and do not expose it
```
joseivanlopez, line 68: Typo: and **does** not expose ...

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,187 @@
...
+      # @param config [Config]
+      def read_elements(config)
+        config.attach(groups)
+        config.attach(users(config))
+      end
+
+      # Returns the list of users
+      #
+      # @return [Array<User>] the collection of users
+      def users(config)
```
joseivanlopez, line 95: NP: I would document `config` param too.

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,187 @@
...
+        all_local = Yast::Users.GetGroups("cn", "local").values +
+          Yast::Users.GetGroups("cn", "system").values
+        all_local.map { |u| group(u) }
+      end
+
+      # Creates a {User} object based on the data structure of a Users user
+      #
+      # @param user_attrs [Hash] a user representation in the format used by Users
+      # @return [User]
+      def user(user_attrs, config)
```
joseivanlopez, line 111: NP: I would document `config` param too.

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,187 @@
...
+      # @param user [Hash] a user or group representation in the format used by Users
+      # @return [String, nil] nil if the value is missing or empty
+      def attr_value(attr, user)
+        value = user[attr.to_s]
+        return nil if value == ""
+
+        value
+      end
+
+      # Creates a {Password} object based on the data structure of a UsersSimple user
```
joseivanlopez, line 166: I guess you meant *Users* instead of *UsersSimple*.

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,187 @@
...
+      #
+      # @param user [Hash] a user representation in the format used by UsersSimple
```
joseivanlopez, line 168: And the same here.

joseivanlopez:
I have a question. How do we obtain both configs, probed and staging? I expected this reader to be able to return both configs. What is the strategic?

src/lib/y2users/users_module/reader.rb
```diff
@@ -0,0 +1,187 @@
...
+        user.home  = attr_value(:homeDirectory, user_attrs)
+        user.gecos = [attr_value(:cn, user_attrs), *user_attrs["addit_data"].split(",")].compact
+        user.system = true if !user.uid && user_attrs["type"] == "system"
+
+        # set password only if specified, nil means not touch it
+        user.password = create_password(user_attrs) if user_attrs["userPassword"]
+
+        # set authorized keys only if set in users module
+        user.authorized_keys = user_attrs["authorized_keys"] if user_attrs["authorized_keys"]
+
```
joseivanlopez, line 128: I miss some info for the user:

* Receive system mail?
* Home mode permissions
* Move home content?
* Create empty home?

Such attributes are not provided by `Y2Users::User` yet, but we are going to add them as part of https://github.com/yast/yast-users/pull/329.

dgdavid: Not sure if two latest can be provided by the reader. Maybe I'm missing something.

src/lib/y2users/users_module/reader.rb
jreidinger: is it usrkel or usrskel?

src/lib/y2users/users_module/reader.rb
jreidinger: Well, if I do that testing Users.Read and export it, then password is not set/read. Probably related to huge UI orientation of Users module.

src/lib/y2users/users_module/reader.rb
jreidinger: for consistency it makes sense.

src/lib/y2users/users_module/reader.rb
jreidinger: I plan to add it when User start having it, otherwise test suite failing.

jreidinger:
@joseivanlopez regarding probed and staging it is much more tricky, as users basically holds just one and remember which ones are modified and which not. So for me it makes the most sense to read probed by linux reader and then get staging from Users as UI working on users on running system.

joseivanlopez:
> @joseivanlopez regarding probed and staging it is much more tricky, as users basically holds just one and remember which ones are modified and which not. So for me it makes the most sense to read probed by linux reader and then get staging from Users as UI working on users on running system.

I am not sure whether using `Y2Users::Linux::Reader` for probed and `Y2Users::UsersModule::Reader` for staging is a good idea. Setting the correspondence between probed and staging users would be tricky.

When a user is edited in the UI, does `Yast::Users` remember the original value of the edited attribute? If so, I guess it is still feasible to generate both configs at the moment of writing. Otherwise, I will propose something like:

~~~
Start YaST users client
Yast::Users reads the system and generates probed config in ConfigManager
UI makes changes in the users
Yast::Users writes by generating the system config in ConfigManager and then calling to Linux::Writer
~~~ 

Do you think this could work?

src/lib/y2users/users_module/reader.rb
jreidinger: and it is read in useradd_config but not in Users module.

dgdavid: > is it usrkel or usrskel?

Second one. Comment updated.

src/lib/y2users/users_module/reader.rb
jreidinger: @dgdavid see testing data in tests. at least empty home aka no skeleton can be obtained from users.

jreidinger:
@joseivanlopez Well, I think users does not store original one, but the way with double calling reader can work.

joseivanlopez (approved):
LGTM

joseivanlopez:
> LGTM, but unit tests are failing.

Ok, only coveralls, false alarm.

PR #329, opened 2021-09-16
Edit users

joseivanlopez:
## Problem

The new `Y2Users` classes support creating and editing users. Right now, such classes are only used for (auto)installation and firstboot. That new classes use the shadow tools to probe and commit users. We plan start using them for the YaST users client instead of the `Yast::Users` module. But the users client allows to manipulate more user attributes than the currently supported by `Y2Users`. So, `Y2Users` needs to be extended in order to support all the use cases provided by YaST users client.

Internal links:

* https://trello.com/c/44pVMU58/2656-5-modifying-users-with-y2users
* https://trello.com/c/qfPotJsP/2649-8-missing-attributes-for-interactive-execution
* https://jira.suse.com/browse/SLE-20592

## Solution

`Y2Users` classes were extended to support all the required attributes for creating and editing users. Moreover, now the commit process can be addressed thanks to a new `CommitConfig` class. For example, such a class allows to indicate whether to move the content of the current home directory.

With this PR, a user can be created and edited completely. Deleting users is still missing in `Y2Users`.

NOTE 1: This PR is part of the work accumulated in the *y2users* branch. 

NOTE 2:  The shadow tools does not support some use cases that YaST users client does:
* Create a home directory for an existing user that has no home.
* Create a new home without skel files. 

The shadow tools should be extended to support that use cases. YaST should provide nothing that the shadow tools does not support.

## Testing

* Manually tested
* Added unit tests.

### DIFF
--- a/doc/attributes.md
+++ b/doc/attributes.md
@@ -8,48 +8,22 @@ The YaST UI for creating and editing a user offers the following attributes:
 
 | UI Field | Y2Users attr | Only when adding | Only when editing | Description |
 | :--- | :--- | :--- | :--- | :--- |
-| full name | gecos | | | |
-| name | name | | | |
-| password | password | | | |
-| system mail | | | | |
+| full name | `#gecos` | | | |
+| name | `#name` | | | |
+| password | `#password` | | | |
+| system mail | `#receive_system_mail` | | | See `MailAliases` module |
 | disable login | * password starting by `!` | | | |
-| uid | uid | | | |
-| home dir | home | | | | |
-| home dir permission | | | | |
-| empty home | | yes | | |
-| Move to new location | | | yes |  |
-| Btrfs subvolume | btrfs_subvolume_home | | | |
-| Additional info | gecos | | | |
-| login shell | shell | | | |
-| group | primary_group | | | |
-| additional groups | groups | | | |
-| ssh public keys | authorized_keys | | | |
-
-### Home management
-
-### Creating a user
-
-* A user can be created with or without home.
-* To create a user without home, the *home dir* path should be set to empty.
-* The home can be created with or without default content (*empty home* checkbox).
-* The home can be created as Btrfs subvolume (*Btrfs subvol* checkbox).
-
-### Editing a user
-
-* If the home path is changed, then a new home is created with the default content.
-* There is no way to create a new home without content (*empty home* checkbox is not available).
-* If the *Move to new location* checkbox is marked, then the content of the old home is moved to the new one and the old home is removed.
-* If the *Move to new location* checkbox is not marked, then the old home is kept.
-* If the old home was a directory, then the new home is created as a directory again.
-* If the old home was a subvolume, then the new home is created as a subvolume again.
-* Side effects of *Move to new location* checkbox:
-  * If the old home was a subvolume and *Move to new location* is not checked, then the new home is created as a directory instead of a subvolume.
-* If the home path is removed, then the home is removed from the user but the home itself (directory or subvolume) is kept.
-* There is no way to remove the home directory/subvolume.
-
-### Deleting a user
-
-* The home is removed too.
+| uid | `#uid` | | | |
+| home dir | `Home#path` | | | | |
+| home dir permission | `Home#permissions` | yes | | `useradd -K HOME_MODE=0755` |
+| empty home | `CommitConfig#home_without_skel` | yes | | `rm -rf` after creating. No way to ignore `/usr/etc/skel` |
+| Move to new location | `CommitConfig#move_home` | | yes |  |
+| Btrfs subvolume | `Home#btrfs_subvol` | | | |
+| Additional info | `#gecos` | | | |
+| login shell | `#shell` | | | |
+| group | `#primary_group` | | | |
+| additional groups | `#groups` | | | |
+| ssh public keys | `#authorized_keys` | | | |
 
 
 ## Password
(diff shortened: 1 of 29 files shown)


doc/attributes.md
```diff
@@ -11,12 +11,12 @@ The YaST UI for creating and editing a user offers the following attributes:
...
 | password | password | | | |
-| system mail | | | | |
+| system mail | receive_system_mail | | | See `MailAliases` module |
 | disable login | * password starting by `!` | | | |
 | uid | uid | | | |
 | home dir | home | | | | |
-| home dir permission | | | | |
-| empty home | | yes | | |
+| home dir permission | umask | yes | | `useradd -K UMASK=666` |
+| empty home | | yes | | `rm -rf` after creating. No way to ignore `/usr/etc/skel` |
```
dgdavid, line 19: Maybe `rm -rf` is not enough. See below screenshot

Screenshot from 2021-09-16 16-18-33


And also https://unix.stackexchange.com/questions/77127/rm-rf-all-files-and-all-hidden-files-without-error/77196

dgdavid: A possible Ruby way, 

```
require 'pathname'

Pathname.new("/home/rmtesting").children.each { |f| f.rm_rf }
```

src/lib/y2users/linux/users_writer.rb
```diff
@@ -217,9 +225,24 @@ def clear_home(user, issues)
         # Pathname.new(user.home).children.each { |f| f.rm_rf }
       rescue Cheetah::ExecutionFailed => e
         issues << Y2Issues::Issue.new(
-          format(_("Something was wrong while cleaning up '%s'"), user.home)
+          format(_("Cannot clean up '%s'"), user.home)
```
dgdavid, line 228: `user.home&.path` or redefine Y2Users::Home#to_s

src/lib/y2users/linux/users_writer.rb
```diff
@@ -217,9 +225,24 @@ def clear_home(user, issues)
...
+      end
+
+      def chown_home(user, issues)
+        return unless user.home&.path
+        return unless File.exist?(user.home.path)
+
+        owner = user.uid
+        owner << ":#{user.gid}" if user.gid
+
+        Yast::Execute.on_target!("/usr/bin/chown", "-R", owner, "\"#{user.home.path}\"")
```
dgdavid, line 240: For consistency with rest of executed commands, I'd use a private constant for chown path.

src/lib/y2users/linux/users_writer.rb
```diff
@@ -435,11 +467,15 @@ def sorted_groups(user)
...
+      # Note that useradd command will not try to create the home directory if it already exists, it
+      # does not matter whether --create-home was passed.
+      #
       # @param user [User]
       # @return [Array<String>]
       def create_home_options(user)
         opts = ["--create-home"]
-        opts << "--btrfs-subvolume-home" if user.btrfs_subvolume_home
+        opts << "--btrfs-subvolume-home" if user.home&.btrfs_subvol?
+        opts << "-K HOME_MODE=#{user.home.permissions}" if user.home&.permissions
```
dgdavid, line 478: Good catch! It's better placed here :ok_hand:

src/lib/y2users/users_module/commit_config_reader.rb
```diff
@@ -29,7 +29,7 @@ module UsersModule
     class CommitConfigReader
       # Generates a collection of commit configs with the information from YaST::Users module
       #
-      # @return [ConfigConfig]
+      # @return [ConfigConfigCollection]
```
dgdavid, line 32: CommitConfigCollection

src/lib/y2users/user.rb
```diff
@@ -100,7 +100,7 @@ class User < ConfigElement
     # Only relevant attributes are compared. For example, the config in which the user is attached
     # and the internal user id are not considered.
     eql_attr :name, :uid, :gid, :shell, :home, :gecos, :source, :password, :authorized_keys,
-      :receive_system_mail, :secondary_groups_name
+      :receive_system_mail, :secondary_groups_name, :receive_system_mail
```
joseivanlopez, line 103: duplicated

src/lib/y2users/linux/writer.rb
```diff
@@ -90,7 +90,7 @@ class Writer
       def initialize(config, initial_config, commit_configs = nil)
         @config = config
         @initial_config = initial_config
-        @commit_configs = commit_configs || CommitConfigCollection.new
+        @commit_configs = commit_configs || Y2Users::CommitConfigCollection.new
```
joseivanlopez, line 93: I think this is not needed.

src/lib/y2users/linux/users_writer.rb
```diff
@@ -165,7 +166,7 @@ def edit_users
 
       # @return [CommitConfig]
       def commit_config(user)
-        commit_configs.by_username(user.name) || CommitConfig.new
+        commit_configs.by_username(user.name) || Y2Users::CommitConfig.new
```
joseivanlopez: I think this is not needed.

src/lib/y2users/linux/users_writer.rb
```diff
@@ -196,8 +189,8 @@ def add_user(user, issues)
       # @param issues [Y2Issues::List] a collection for adding an issue if something goes wrong
       def create_user(user, issues)
         try_create_user(user, issues)
+        root_aliases << user.name if user.receive_system_mail
```
joseivanlopez, line 192: NP: maybe we should rename this method as `#receive_system_mail?`

doc/attributes.md
```diff
@@ -8,35 +8,55 @@ The YaST UI for creating and editing a user offers the following attributes:
...
-| additional groups | groups | | | |
-| ssh public keys | authorized_keys | | | |
-
-### Home management
-
-### Creating a user
+| uid | `#uid` | | | |
+| home dir | `Home#path` | | | | |
+| home dir permission | `Home#permissions` | yes | | `useradd -K HOME_MODE=0755` |
+| empty home | `CommitConfig#home_without_skel` | yes | | `rm -rf` after creating. No way to ignore `/usr/etc/skel` |
```
jreidinger, line 19: actually there is a way. Using `--system` but it has different consequences.

doc/attributes.md
```diff
@@ -8,35 +8,55 @@ The YaST UI for creating and editing a user offers the following attributes:
...
+| Move to new location | `CommitConfig#move_home` | | yes |  |
+| Btrfs subvolume | `Home#btrfs_subvol` | | | |
```
jreidinger, line 21: this only when adding and only when edditing is kind of strange for me, as it should be exclusive. What having just one collumn having there Add/Edit/All ?

doc/attributes.md
joseivanlopez: yes :(

ancorgs (approved):
LGTM, despite the broken continuous integration.

ancorgs (approved):
Reapproved

PR #330, opened 2021-09-23
use y2users for local users and groups

jreidinger:


src/include/users/complex.rb
```diff
@@ -26,6 +26,13 @@
 #		Jiri Suchomel <jsuchome@suse.cz>
 #
 # $Id$
+
```
joseivanlopez, line 29: I would not include all these changes related to the integration of Y2Users and Yast::Users. I see this is useful for testing, but there is a separate PR for such an integration.

src/modules/Users.pm
```diff
@@ -3986,6 +3996,7 @@ sub WriteCustomSets {
 ##------------------------------------
 # Writes settings to /etc/defaults/useradd
 sub WriteLoginDefaults {
+    return 1;
```
joseivanlopez, line 3999: The same here. I would avoid all changes related to the integration.

src/lib/y2users/users_module/reader.rb
```diff
@@ -33,15 +33,17 @@ module Y2Users
   module UsersModule
     # Class for reading the configuration from Yast::Users module
     class Reader
-      # Generates a new config with the information from YaST::Users module
+      # Generates a new configs with the information from YaST::Users module
       #
-      # @return [Config]
+      # @return [Config,Config] returns pair of configs. system one and modified one
```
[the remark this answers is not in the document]
jreidinger: well, I want to emphasize that it is pair and not just random array. Sadly I do not find any example of such yardoc tag.

src/include/users/complex.rb
jreidinger: Well, my goal is to make y2users branch working and @ancorgs then adapt it to whatever style fits. Also it is useful that it reports now issues, which really helps with feedback and testing.

PR #332, opened 2021-09-29
Fix clients

joseivanlopez:
## Problem

Clients for creating a new user during installation and firstboot rely on the `Yast::InstUserFirstDialog`. Such a dialog receives a user and performs changes over it according to the values entered in the dialog. But the user home is not assigned at all. Because of that, the new user created in that clients will not contain a home directory/subvolume (note that `Y2Users::Linux::Writer` only creates a home if `user#home` has a path).

* See https://github.com/yast/yast-firstboot/pull/129

## Solution

Assign the home in the dialog, similar to password and other attributes.

NOTE: This PR is part of the work accumulated in the y2users branch.

## Testing

* Adapted unit test
* Manually tested


src/lib/users/dialogs/inst_user_first.rb
```diff
@@ -425,10 +425,24 @@ def process_new_user_form
       @full_name = UI.QueryWidget(Id(:full_name), :Value)
 
       user.name = @username
+      user.home = Y2Users::Home.new("/home/#{user.name}")
```
jreidinger, line 428: no escaping is needed?

jreidinger: and why is this default here? I kind of expect that user propose this home as default value.

joseivanlopez: I don't think so. According to *useradd* man page:

> Usernames must start with a lower case letter or an underscore, followed by lower case letters, digits, underscores, or dashes. They can end with a dollar sign. In regular expression terms: [a-z_][a-z0-9_-]*[$]?
>
> Usernames may only be up to 32 characters long.

The only way to use other characters for login is with *--badname* option, but we don't use such an option:

> --badname Allow names that do not conform to standards.

joseivanlopez: > and why is this default here? I kind of expect that user propose this home as default value.

This is something we can add to `User`. For now, the user object is initialized without home. Maybe we should add some constructor methods:

# `#new`: with home and system `false.`
* `#create_system`: without home and system `true`.

Anyway, in this dialog we have to recreate the home, because in case of editing a user (when going back in the installation/firstboot), the home path should be adapted to the new user name (just in case the user name was modified).

joseivanlopez: I guess we can also have a method in `User` to obtain the default home or something like that, in order to have the default generation (i.e., "/home/#{user.name}") only in one place.

jreidinger: yeap, exactly my idea. We should have one place for default value for home. if it is spread over multiple clients, it is wrong from my POV

ancorgs: Yes, completely agree. I would have a `User#default_home_path` to return the string and `User#default_home` to return the object.

src/lib/y2users/user.rb
```diff
@@ -134,6 +148,13 @@ def initialize(name)
       @system = false
     end
 
+    # Default home for the user
+    #
+    # @return [Home]
```
jreidinger, line 153: I would document that it creates a new instance, so all previous home related config is lost.

jreidinger (approved):

PR #333, opened 2021-09-30
Yast::Users uses Y2Users for writing local configuration

ancorgs:
This is part of https://jira.suse.com/browse/SLE-20563, whose ultimate goal is to ensure YaST2 uses the tools in the `shadow` package (ie. `useradd` and friends) to modify local and system users, and also to update the local system configuration (eg. the values at `/etc/default/useradd`).

This pull request changes the implementation of `Yast::Users.write`. Instead of modifying the local files by itself, now that method relies on the `Y2Users` classes to change the system. Those classes rely on the `shadow` tools, as desired.

Done some preliminary manual testing and everything seems to work. Changes are indeed performed by the `Y2Users' classes and any error found during the process is returned back to the legacy `Yast::Users` module for it to show the error to the user and manage the situation.

Tested both by pressing "ok" (and thus finalizing the module) and with the "write changes now" button.

Of course, further testing of several possible situations would be desirable.


src/modules/Users.pm
```diff
@@ -4476,149 +4307,24 @@ sub Write {
...
- 	}
-        if (! WriteShadow ()) {
-            $ret = Message->ErrorWritingFile ("$base_directory/shadow");
-	    Report->Error ($ret);
-	    return $ret;
-        }
+	# There used to be a big loop here managing homedir changes, but is not longer necessary:
+	#  - work with homes for LDAP users are ruled in WriteLDAP
+	#  - homes for local and system users are handed by Y2Users
+	#  - is not possible to add NIS users or to configure their homes
```
jreidinger, line 4327: these lines also causing troubles https://github.com/yast/yast-users/pull/333/files#diff-cf4196c4e5f27891c3cbd73e1ae939d14daea245d725677c7b51cdef7b833ad9R4367
it causes that org_user is removed and causes troubles to users reader

ancorgs:
Found a situation that works worse than before.

1. Open yast2-users
2. Change only the list of secondary groups of a user
3. Press "ok" or "Write changes now"
4. With this PR, the writer is not even called so nothing happens. But the yast2-users currently in TW works as expected.

ancorgs:
> Found a situation that works worse than before.

I partially fixed it. Now the writer is executed even if only the group membership was changed. Fix committed via a "commit amend".

But it still does not modify the information in the system, which could be an error in the writer itself or in `UsersModule::Reader`.

src/Makefile.am
```diff
@@ -130,6 +129,11 @@ ylib_y2users_linux_DATA = \
   lib/y2users/linux/reader.rb \
   lib/y2users/linux/writer.rb
 
+ylib_y2users_users_moduledir = @ylibdir@/y2users/users_module
+ylib_y2users_users_module_DATA = \
+  lib/y2users/users_module/reader.rb \
+  lib/y2users/users_module/commit_config_reader.rb
```
jreidinger, line 135: good catch :+1:

src/modules/Users.pm
```diff
@@ -3996,21 +3973,8 @@ sub WriteCustomSets {
...
-	$ret = $ret && SCR->Write (".etc.default.useradd.\"\Q$key\E\"", $value);
-    }
-
-    if ($ret) {
-	SCR->Write (".etc.default.useradd", "force");
-    }
-    y2milestone ("Succesfully written useradd defaults: $ret");
-    y2usernote ("File '/etc/default/useradd' was modified.");
-    return $ret;
+    y2error ("Calling obsolete WriteLoginDefaults");
```
jreidinger, line 3976: is it really called anywhere? As it does not have typedef macro, it cannot be called from yast. Maybe it can be just removed?

src/modules/Users.pm
```diff
@@ -5741,12 +5364,8 @@ sub SetRootPassword {
...
 BEGIN { $TYPEINFO{WriteRootPassword} = ["function", "boolean"];}
 sub WriteRootPassword {
-
-    my $self		= shift;
-    # Crypt the root password according to method defined in encryption_method
-    my $crypted		= $self->CryptPassword ($root_password, "system");
-    Syslog->Log ("Root password changed by YaST");
-    return SCR->Write (".target.passwd.root", $crypted);
+    y2error ("Calling obsolete WriteRootPassword");
+    return 1;
```
jreidinger, line 5368: here it make sense as it is exported to yast

joseivanlopez (approved):
LGTM

PR #335, opened 2021-10-04
Read useradd config

joseivanlopez:
## Problem

The YaST users client has a tab for editing the useradd settings (i.e., */etc/default/useradd*). The current useradd config is read by the *Users* module, which directly parses */etc/default/useradd*. But such a file could not exist in the system, so *Users* module would propose its own default values instead of relying on the default values proposed by the shadow tools (*useradd -D* command). Moreover, some of the values offered by the "Defaults for New Users" tab are not supported by the shadow tools anymore, for example *skel* and *groups*.    

* https://trello.com/c/hWPBYvVY/2651-3-use-y2usersuseraddconfig-from-to-yastusers
* Part of https://jira.suse.com/browse/SLE-20563

## Solution

Unsupported attributes are removed from the UI (*skel* and *groups*), and the *Y2Users::Linux::UseraddConfigReader* is used for reading values from the system. Entered values in the "Defaults for New Users" tab are persisted into the system by *Y2Users::Linux::Writer*.

Screenshot from 2021-10-05 08-50-23

NOTE: This PR is part of the work accumulated in the y2users branch

## Testing

Manually tested.

### DIFF
--- a/src/include/users/widgets.rb
+++ b/src/include/users/widgets.rb
@@ -142,7 +142,7 @@ module Yast
         "samba"    => _("Samba")
       }
 
-      # auth methods and respective module calls 
+      # auth methods and respective module calls
       @auth_methods = {
         "nis" => {
           # menubutton label
@@ -195,10 +195,8 @@ module Yast
               VSpacing(0.2),
               "defaults_global",
               "defaultgroup",
-              "groups",
               "shell",
               HBox("home", VBox(Label(""), "browse_home")),
-              HBox("skel", VBox(Label(""), "browse_skel")),
               "umask",
               "expire",
               "inactive",
@@ -210,12 +208,9 @@ module Yast
           "widget_names" => [
             "defaults_global",
             "defaultgroup",
-            "groups",
             "shell",
             "home",
             "browse_home",
-            "skel",
-            "browse_skel",
             "umask",
             "expire",
             "inactive"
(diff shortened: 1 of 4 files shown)


src/include/users/widgets.rb
```diff
@@ -195,10 +195,8 @@ def initialize_users_widgets(include_target)
               VSpacing(0.2),
               "defaults_global",
               "defaultgroup",
-              "groups",
               "shell",
               HBox("home", VBox(Label(""), "browse_home")),
-              HBox("skel", VBox(Label(""), "browse_skel")),
```
jreidinger, line 201: can you paste to description also screenshot how it looks after that removal? Just to get impression how it looks like.

src/include/users/widgets.rb
```diff
@@ -375,28 +359,6 @@ def initialize_users_widgets(include_target)
...
-            "boolean (string, map)"
-          )
-        },
-        "browse_skel"       => {
-          "widget"  => :push_button,
-          # push button label
-          "label"   => _("Bro&wse..."),
-          "opt"     => [:key_F7],
-          "handle"  => fun_ref(
-            method(:HandleBrowseDirectory),
```
jreidinger, line 395: note for me: handle is used also for browse home so, cannot be removed.

src/include/users/widgets.rb
```diff
@@ -1851,20 +1737,13 @@ def StoreDefaults(key, event)
       event = deep_copy(event)
       defaults = Users.GetLoginDefaults
       new_defaults = {}
-      Builtins.foreach(["home", "shell", "skel", "inactive", "umask"]) do |key2|
+      Builtins.foreach(["home", "shell", "inactive", "umask"]) do |key2|
```
jreidinger, line 1740: when I see that umask, is not that now readonly?

jreidinger (approved):
just NPs

src/include/users/widgets.rb
joseivanlopez: It is not managed by *useradd* anymore. Now, that value is in *login.defs* and it can be read and written.

joseivanlopez: See https://github.com/yast/yast-users/blob/y2users/src/lib/y2users/useradd_config.rb#L108-L123.

PR #338, opened 2021-10-05
Remove home config

jreidinger:


jreidinger:
@ancorgs yes, it should be always there https://github.com/yast/yast-users/blob/master/src/modules/Users.pm#L1892
and also list is always defined https://github.com/yast/yast-users/blob/master/src/modules/Users.pm#L101

ancorgs (approved):
LGTM

PR #339, opened 2021-10-05
Delete users and groups

joseivanlopez:
https://trello.com/c/XcAKT8r0/2657-3-deleting-users-with-y2users

### DIFF
--- a/src/lib/y2users/linux/edit_user_action.rb
+++ b/src/lib/y2users/linux/edit_user_action.rb
@@ -51,13 +51,13 @@ module Y2Users
       def run_action
         options = usermod_options
         Yast::Execute.on_target!(USERMOD, *options, initial_user.name) if options.any?
-        true
+        result(true)
       rescue Cheetah::ExecutionFailed => e
         issues << Y2Issues::Issue.new(
           format(_("The user '%{username}' could not be modified"), username: initial_user.name)
         )
         log.error("Error modifying user '#{initial_user.name}' - #{e.message}")
-        false
+        result(false)
       end
 
       # Command to modify the user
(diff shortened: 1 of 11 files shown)


src/lib/y2users/linux/edit_user_action.rb
```diff
@@ -0,0 +1,124 @@
...
+        log.error("Error modifying user '#{initial_user.name}' - #{e.message}")
+        false
+      end
+
+      # Command to modify the user
+      #
+      # @return [Array<String>] usermod options
+      # rubocop:disable Metrics/CyclomaticComplexity
+      # rubocop:disable Metrics/AbcSize
+      # rubocop:disable Metrics/PerceivedComplexity
```
jreidinger, line 68: is those still needed?

src/lib/y2users/linux/remove_home_content_action.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      # Command for finding files
+      FIND = "/usr/bin/find".freeze
+      private_constant :FIND
+
+      # Clear the content of the home directory/subvolume for the given user
+      #
+      # Issues are generated when the home cannot be cleaned up.
+      def run_action
+        Yast::Execute.on_target!(FIND, user.home.path, "-mindepth", "1", "-delete")
```
jreidinger, line 48: does find also find hidden files? and what about removing directory?

src/lib/y2users/linux/users_writer.rb
```diff
@@ -61,11 +59,14 @@ def initialize(config, initial_config, commit_configs)
...
-        with_issues do |issues|
-          issues.concat(add_users)
-          issues.concat(edit_users)
-          issues.concat(update_root_aliases)
-        end
+        @issues = Y2Issues::List.new
+
+        add_users
+        edit_users
+        # delete_users
```
jreidinger, line 66: order is wrong, you have to first delete and then add/edit to prevent name/home collisions.

PR #340, opened 2021-10-06
Fix groups

jreidinger:
Issue: difference between old and new users in group is not detectable.

Fix: ensure it creates original group even when just syncing from edit user. And also ensure it do deep clone.


ancorgs:
The change makes sense, but at first glance I don't see why this works in the master branch without this fix.

ancorgs:
@jreidinger got it. Makes sense, the version in master is "configuration-file-driven". It regenerates `/etc/passwd` and/or `/etc/group` as a whole when something changes and I guess there are many ways in which such a rewrite can be triggered.

ancorgs (approved):
LGTM

PR #342, opened 2021-10-07
Stop exporting (and using in general) obsolete useradd configuration attributes

ancorgs:
This is part of the big modification we are doing in the separate `y2users` branch in order to ensure the interactive module of YaST2 uses the `shadow` tools under the hood.

## The problem

The useradd configuration managed by YaST used to contain obsolete attributes that are not longer useful. AutoYaST ignores them (see more details at https://github.com/SUSE/doc-sle/pull/901), so it makes no sense to export them.

After the recent changes in this branch, those useless attributes are also ignored when using YaST interactively and are not longer displayed in the UI.

## Solution

This pull request finally removes those attributes from the code (except in the very special case of LDAP), ensuring they are not longer exported when the system is cloned by AutoYaST.

It also ensures that an AutoYaST profile that includes obsolete attributes does not pollute the internal data structures.

### DIFF
--- a/src/modules/Y2UsersLinux.rb
+++ b/src/modules/Y2UsersLinux.rb
@@ -30,10 +30,17 @@ module Yast
   class Y2UsersLinuxClass < Module
     include Yast::Logger
 
-    EXPORTED_USERADD_ATTRS = [
-      "group", "home", "shell", "skel", "umask", { "inactivity_period" => "inactive" },
-      { "expiration" => "expire" }
-    ]
+    # Mapping from {Y2Users::UseraddConfig} attributes into names of the keys used in the
+    # {Yast::Users} hashes
+    EXPORTED_USERADD_ATTRS = {
+      group:             "group",
+      home:              "home",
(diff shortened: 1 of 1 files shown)


jreidinger (approved):

src/modules/Y2UsersLinux.rb
```diff
@@ -30,20 +30,27 @@ module Yast
   class Y2UsersLinuxClass < Module
     include Yast::Logger
 
-    EXPORTED_USERADD_ATTRS =
-      ["group", "home", "inactivity_period", "expiration", "shell", "skel", "umask"].freeze
+    EXPORTED_USERADD_ATTRS = [
+      "group", "home", "shell", "skel", "umask", { "inactivity_period" => "inactive" },
+      { "expiration" => "expire" }
```
jreidinger, line 35: this is kind of tricky structure. I think it should be either documented or changed to hash that is mapping ( even if there is same key and value )

jreidinger (approved):

PR #345, opened 2021-10-18
Do not crash because missing configuration files

dgdavid:
We have got a bug report about new code base crashing due to a missing /etc/group file. It happens because the reader was not checking the file existence before trying to read it.

* https://trello.com/c/yfe8ZNVF/
* https://bugzilla.suse.com/show_bug.cgi?id=1191591

To fix it, the reader now simply checks if  `File.exist?`, returning a no-data content and logging an error message if not.

Additionally, a new helper method has been introduced to get rid of /etc/default/useradd fixture file because it was dropped from the shadow package too (see https://bugzilla.suse.com/show_bug.cgi?id=1191189#c2)


joseivanlopez (approved):
LGTM

PR #349, opened 2021-10-19
Use shadow tools in users client

joseivanlopez:
Adapts YaST users client to use the shadow tools for performing changes into the system.

* https://jira.suse.com/browse/SLE-20592

NOTE: this PR is only about merging the accumulated work. Everything has been already reviewed and manually tested.


ancorgs (approved):
LGTM. As mentioned, all individuals commit should already be reviewed, so I only checked the changes to the `packages` directory.

PR #350, opened 2021-10-27
Read home permissions and root aliases

dgdavid:
Adapt Linux readers to read home permissions and root aliases too.

* https://trello.com/c/UIMO6uJg/2655-1-missing-atributes-on-linuxreader

### DIFF
--- a/src/lib/y2users/linux/base_reader.rb
+++ b/src/lib/y2users/linux/base_reader.rb
@@ -20,14 +20,9 @@
 require "yast"
 require "abstract_method"
 require "y2users/config"
-require "y2users/login_config"
 require "y2users/parsers/group"
 require "y2users/parsers/passwd"
 require "y2users/parsers/shadow"
-require "y2users/linux/useradd_config_reader"
-require "users/ssh_authorized_keyring"
-
-Yast.import "Autologin"
 
 module Y2Users
   module Linux
--- a/src/lib/y2users/linux/base_reader.rb
+++ b/src/lib/y2users/linux/base_reader.rb
@@ -28,6 +28,7 @@ require "y2users/linux/useradd_config_reader"
 require "users/ssh_authorized_keyring"
 
 Yast.import "Autologin"
+Yast.import "MailAliases"
 
 module Y2Users
   module Linux
@@ -41,6 +42,7 @@ module Y2Users
       def read
         Config.new.tap do |config|
           read_elements(config)
+          read_root_aliases(config)
           read_passwords(config)
           read_authorized_keys(config)
           read_useradd_config(config)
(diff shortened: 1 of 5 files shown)


src/lib/y2users/linux/base_reader.rb
```diff
@@ -42,9 +42,6 @@ def read
         Config.new.tap do |config|
           read_elements(config)
           read_passwords(config)
-          read_authorized_keys(config)
```
joseivanlopez, line 45: Please, remove the unneeded requires (i.e, login_config, useradd_config_reader and ssh_authorized_keyring).

src/lib/y2users/linux/reader.rb
```diff
@@ -20,11 +20,93 @@
...
+      #
+      # @return [Array<Y2Users::User>]
+      def read_home_permissions(config)
+        config.users.reject(&:system?).each do |user|
+          next unless user.home && Dir.exist?(user.home.path)
+
+          user.home.permissions = Yast::Execute.on_target!(
+            STAT, "--printf", "%a", user.home.path,
+            stdout: :capture
+          ).prepend("0")
```
joseivanlopez, line 73: Please, add a comment to explain the result of start %a and why it is needed to prepend a 0.

src/lib/y2users/linux/base_reader.rb
dgdavid: Arrrrgghhh! :man_facepalming: :man_facepalming: :man_facepalming:

src/lib/y2users/linux/reader.rb
dgdavid: While trying to write down _why it is needed_ I realized that using the [`#` printf flag character](https://www.man7.org/linux/man-pages/man3/printf.3.html) it's better. Honestly, [`stat`documentation](https://github.com/coreutils/coreutils/blob/a0468760ffa9f7843bbd0d6cf60685f63d33b600/src/stat.c#L1783) was not so clear to me the first time I read it.

Please, check the added comment and suggest changes if you find it confusing.

Thanks!

src/lib/y2users/linux/reader.rb
```diff
@@ -19,12 +19,101 @@
...
+      private_constant :STAT
+
+      # Reads home permissions
+      #
+      # @return [Array<Y2Users::User>]
+      def read_home_permissions(config)
+        config.users.reject(&:system?).each do |user|
+          next unless user.home && Dir.exist?(user.home.path)
+
+          # NOTE: '#' printf flag is needed to make sure permissions bits
```
joseivanlopez, line 74: I would simply this:

~~~
# "stat --printf %#a" returns the permissions in octal format (e.g., 0755), as expected by {Home#permissions}. 
~~~

joseivanlopez (approved):
LGTM

PR #352, opened 2021-11-03
Fix check for user home existence

dgdavid:
## Problem

When installing via an AutoYaST profile setting [`<authorized_keys>`](https://doc.opensuse.org/projects/autoyast/#id-1.9.5.2.34.3.9.14) for users the system is installed without provided SSH public keys, as reported in http://bugzilla.suse.com/show_bug.cgi?id=1192178.

The problem is in the [check for user home existence](https://github.com/yast/yast-users/blob/f6d4a583ee13c16201145bf68e2f2028e55d9e76/src/lib/y2users/linux/users_writer.rb#L328-L332), which is looking at `/home/<username>` instead `/mnt/home<username>` during installation.

## Solution

~~Use `Yast::Execute.on_target!` and `stat` from [GNU core utils](https://github.com/coreutils/coreutils) command instead of [File.exist?](https://ruby-doc.org/core-3.0.2/File.html#method-c-exist-3F) Ruby method.~~

Use [Yast::FileUtils.IsDirectory](https://github.com/yast/yast-yast2/blob/25a8bb990c4b9e81d2af2c4a699288dd95f90900/library/general/src/modules/FileUtils.rb#L61-L76) instead of [File.exist?](https://ruby-doc.org/core-3.0.2/File.html#method-c-exist-3F) Ruby method. (see https://github.com/yast/yast-users/pull/352#discussion_r741873179)

## Tests

* Unit test adapted
* Tested manually

### DIFF
--- a/src/lib/y2users/linux/users_writer.rb
+++ b/src/lib/y2users/linux/users_writer.rb
@@ -328,7 +328,10 @@ module Y2Users
       def exist_user_home?(user)
         return false unless user.home&.path
 
-        File.exist?(user.home.path)
+        Yast::Execute.on_target!("/usr/bin/stat", user.home.path)
+        true
+      rescue Cheetah::ExecutionFailed
+        false
       end
     end
   end
--- a/src/lib/y2users/linux/users_writer.rb
+++ b/src/lib/y2users/linux/users_writer.rb
@@ -328,10 +328,7 @@ module Y2Users
       def exist_user_home?(user)
         return false unless user.home&.path
 
-        Yast::Execute.on_target!("/usr/bin/stat", user.home.path)
-        true
-      rescue Cheetah::ExecutionFailed
-        false
+        Yast::FileUtils.IsDirectory(user.home.path)
       end
     end
   end
--- a/test/lib/y2users/linux/users_writer_test.rb
+++ b/test/lib/y2users/linux/users_writer_test.rb
@@ -173,7 +173,8 @@ describe Y2Users::Linux::UsersWriter do
         before do
           target_user.home.path = "/home/new_home"
 
-          allow(subject).to receive(:exist_user_home?).with(target_user).and_return(true)
+          allow(Yast::FileUtils).to receive(:IsDirectory).with(target_user.home.path)
+            .and_return(true)
         end
 
         let(:commit_config) do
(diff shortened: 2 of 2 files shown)


src/lib/y2users/linux/users_writer.rb
```diff
@@ -328,7 +328,10 @@ def commit_config(user)
       def exist_user_home?(user)
         return false unless user.home&.path
 
-        File.exist?(user.home.path)
+        Yast::Execute.on_target!("/usr/bin/stat", user.home.path)
+        true
```
dgdavid, line 332: Another option would be 

```suggestion
        Yast::Execute.on_target!("/usr/bin/stat", "--printf", "%F", user.home.path, stdout: :capture) == "directory"
```

but I have doubts if it is a good idea. What do you think?

dgdavid: BTW, the use of `Yast::Execute` instead of `Yast::FileUtils.Exists?` is intentional, after discussed it with @joseivanlopez

ancorgs: I prefer the suggested version with "directory".

And please add a comment to the code explaining why you decided to not use `Yast::FileUtils.Exists?` "After discussing it with Iván" is a bit vague. ;-)

ancorgs: That being said, I think the most YaSTie way for checking for a file/directory in the target system is using the target agent: https://github.com/yast/yast-core/blob/master/doc/systemagent.md

dgdavid: Thanks for the comment @ancorgs. Yes, it was a bit vague from my side to not give a more elaborated explanation :wink: Summarizing we (specially me) thought that `Yast::Execute` somehow obsoletes SCR target agents. But as you explained in the YaST  IRC channel:

> **\<ancorgs\>** to a certain extend, Yast::Execute substitutes/obsoletes target.bash and target.bash_output... but not the rest of the utilities in the target agent

So, let's change it and use `Yast::FileUtils.Exists?`.

dgdavid: > let's change it and use `Yast::FileUtils.Exists`

Or even [`Yast::FileUtils.IsDirectory`](https://github.com/yast/yast-yast2/blob/25a8bb990c4b9e81d2af2c4a699288dd95f90900/library/general/src/modules/FileUtils.rb#L61-L76)

ancorgs (approved):
LGTM

PR #353, opened 2021-11-08
Do not support group passwords

dgdavid:
## Problem

The [not so widespread](https://unix.stackexchange.com/questions/93123/typical-use-case-for-a-group-password) _passworded_ group feature is considered

> _an inherent security problem since more than one person is permitted to know the password._ — https://www.man7.org/linux/man-pages/man1/gpasswd.1.html

Thus, the support for it in yast2-users will be dropped in the [new code base relying on shadow tools](https://github.com/yast/yast-users/pull/349).

## Solution

Stop supporting it in the new yast2-users code base, for which the below actions have been taken

* Do not export either, `<group_password>` or `<encrypted>` tags to the AutoYaST profile.
* Do not look for those values while importing groups.
* Be compatible with old generated profiles holding group passwords by creating the group **without** password AND warning the user about it (see screenshot below).

## Screenshot

Screenshot_opensusetumbleweed-2_2021-11-08_09:38:13

<p align="center"><em>A warning during an AutoYaST installation using a profile with a passworded group.</em></p>

## Tests 

* Added unit tests
* Tested manually both, during installation (import) and in a running system (export)

## Related PR

* https://github.com/SUSE/doc-sle/pull/1003

## Other links

* https://trello.com/c/lPsu4nb7/ (internal link).

### DIFF
--- a/src/autoyast-rnc/users.rnc
+++ b/src/autoyast-rnc/users.rnc
@@ -19,6 +19,9 @@ groups =
     LIST,
     gr_group*
   }
+
+# TODO: group_password and encrypted are no longer supported by YaST. We should
+# remove them from the schema at some point.
 gr_group = element group {
   MAP,
   (
--- a/src/autoyast-rnc/users.rnc
+++ b/src/autoyast-rnc/users.rnc
@@ -20,8 +20,8 @@ groups =
     gr_group*
   }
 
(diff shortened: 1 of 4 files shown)


joseivanlopez (approved):
LGTM

PR #355, opened 2021-11-10
Always create home

joseivanlopez:
## Problem

In the *YaST users* client, leaving the "Home Directory" field as empty leads to a weird behavior. The old code really left the field empty in */etc/passwd* and skipped the creation of the directory. But that behavior was probably fully unintended since the empty "Home Directory" field was automatically re-filled with the default data on every tab switch. Note that leaving the */etc/passwd* field empty doesn't seem to be fully correct - the possibility is not mentioned in the man page for */etc/passwd* and `su` complains saying the entry is incomplete.

Currently, `Y2Users::Linux::Writer` does not create the home directory for a local user if the home path is unknown. But `useradd` always adds the default home path to */etc/passwd*. Because of that, the current behaviour of *YaST users* client is slightly different when the "Home Directory" is set to empty.

Moreover, AutoYaST is expected to create the default home directory when no home path is indicated in the profile. Right now, the home is only created if the path is known. 

* https://trello.com/c/E3wYL0ob/2723-2-do-not-allow-home-directory-to-be-empty
* https://trello.com/c/l5mmMzXQ/2735-fix-autoyast-import-of-home-dir

## Solution

Now, the default home directory is always created for local users when the home path is unknown. Moreover, *YaST users* validates the "Home Directory" in order to avoid setting it as empty. In that case, the user is alerted with a popup and the previous value is restored.

https://user-images.githubusercontent.com/1112304/141127881-a505892f-5ada-48a0-a542-bfcb247d1338.mp4

### Testing

* Added unit tests.
* Manually tested.

### DIFF
--- a/src/include/users/dialogs.rb
+++ b/src/include/users/dialogs.rb
@@ -1418,6 +1418,12 @@ module Yast
               focus_tab.call(current, :home)
               next
             end
+            if new_home.empty?
+              Report.Error(_("Home cannot be empty."))
+              UI.ChangeWidget(Id(:home), :Value, home)
+              focus_tab.call(current, :home)
+              next
+            end
             failed = false
             begin
               error_map = Users.CheckHomeUI(new_i_uid, new_home, ui_map)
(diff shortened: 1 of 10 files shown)


dgdavid (approved):
LGTM

PR #356, opened 2021-11-11
Fix UsersModule::Reader

joseivanlopez:
## Problem

There are issues when a user is added in the "YaST client" and the option "Write Changes Now" is used to save the changes into the system:
* The next time that `Linux::Writer` is called (i.e., by clicking "Ok" or "Write Changes Now"), it tries to create the new users and groups again.
* When editing a group to add a user to its list, the writer complains if the user does not exist in the system yet (i.e., the user is new).

## Solution

The detection of new users and groups by `UsersModule::Reader` was wrong. It was based on the *what* attribute, which usually contains "add_user" or "add_group" when the user/group is new. But that is not always true. In some cases, for example when a user is added to a group, the "what" attribute contains other values like "group_change". Because of that, the `UsersModule::Reader` could consider new users as existing ones.

Moreover, after using  "Write Changes Now", the "what" attribute is not refreshed, so the new users and groups are still considered as new elements the next time that changes are written to the system.

The solution was to use the "modified" attribute instead of "what". The "modified" attribute always contains "added" when the element is new. After using "Write Changes Now", the attribute is properly updated in order to not consider the element as new again.

## Testing

* Adapted unit tests.
* Manually tested.

### DIFF
--- a/src/lib/y2users/users_module/reader.rb
+++ b/src/lib/y2users/users_module/reader.rb
@@ -104,7 +104,7 @@ module Y2Users
         all_local = Yast::Users.GetUsers("uid", "local").values +
           Yast::Users.GetUsers("uid", "system").values
         all_local.each do |user_hash|
-          if user_hash["what"] == "add_user" # new user
+          if user_hash["modified"] == "added" # new user
             target_config.attach(user(user_hash, target_config))
           else # everything else is edit of different kinds
             # at first create original user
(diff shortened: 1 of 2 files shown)


dgdavid (commented):
I have tested it manually and it works! :partying_face:  Thanks! But maybe it worth extending https://github.com/yast/yast-users/blob/y2users/test/lib/y2users/users_module/reader_test.rb for adding examples where `what` is not _add\_user_ or _add\_group_ but still the element is added. What do you think?

dgdavid (approved):
LGTM

PR #358, opened 2021-11-30
AutoYaST: Improve ignored group passwords

dgdavid:
## Problem

Although YaST [does not support group passwords anymore](https://github.com/yast/yast-users/pull/353) and ignores them when importing the profile, the user receives a warning about it ONLY when the found `<group_password>` does not mean "no password". I.e, when the value is neither, empty nor `x`.

However, other values like `!` and `*` are used to express the absence of the group password too BUT we are warning the user about ignoring such a _no password_. See https://openqa.suse.de/tests/7757865#step/installation/48 (if it has not expired yet)

<table>
  <thead>
    <tr>
      <th>The AutoYaST profile excerpt</th>
      <th>The wrong/not needed warning</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>

   ```xml
    <group>
      <encrypted config:type="boolean">true</encrypted>
      <gid>483</gid>
      <group_password>x</group_password>
      <groupname>systemd-timesync</groupname>
      <userlist/>
    </group>
    <group>
      <encrypted config:type="boolean">true</encrypted>
      <gid>498</gid>
      <group_password>!</group_password>
      <groupname>mail</groupname>
      <userlist>postfix</userlist>
    </group>
    <group>
      <encrypted config:type="boolean">true</encrypted>
      <gid>476</gid>
      <group_password>x</group_password>
      <groupname>vnc</groupname>
      <userlist/>
    </group>
   ```
   </td>
   <td>

  Not needed warning

   </td>
    </tr>
  </tbody>
</table>

## Solution

Group all of them on a list of _blank_ password values to be ignored.

## Tests

  * Unit tests updated.

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Nov 30 20:40:58 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- AutoYaST: do not warning about deprecated <group_password>
+  when found a blank one (related to jsc#SLE-20592).
+- 4.4.9
+
+-------------------------------------------------------------------
 Tue Oct 19 10:36:08 UTC 2021 - José Iván López González <jlopez@suse.com>
 
 - Users client uses the shadow tools to write changes into the
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -2,7 +2,7 @@
 Tue Nov 30 20:40:58 UTC 2021 - David Diaz <dgonzalez@suse.com>
 
 - AutoYaST: do not warning about deprecated <group_password>
-  when found a blank one (related to jsc#SLE-20592).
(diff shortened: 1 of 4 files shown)


package/yast2-users.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Nov 30 20:40:58 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- AutoYaST: do not warning about deprecated <group_password>
+  when found a blank one (related to jsc#SLE-20592).
```
jreidinger, line 5: this is not precise. I would mention here explicitelly that we add "!" and "*" to ignored ones.

dgdavid: Let's improve it

jreidinger (approved):

PR #361, opened 2022-03-22
Fix import users

joseivanlopez:
## Problem

Now, YaST relys on the shadow tools for creating users and groups. The `useradd` command fails if the given *gid* does not exist. This situation could happen when importing users from a previous installation. It worked with the old code because *gid* was sanitized before writing the *passwd* file, falling back to a default *gid*.

* https://bugzilla.suse.com/show_bug.cgi?id=1197040
* https://trello.com/c/O0yWuGtg/2787-2-importing-a-user-with-unknown-gid

Screenshot_deeping-test_2021-11-10_16_15_48

## Solution

Remove *gid* from the imported users when such a group is not present in the configuration. Note that `useradd` will assign the default group according to the */etc/default/useradd* config.

NOTE: this PR will be merged into *pre-SLE-15-SP4* branch. That branch will be merged into master after branching SLE-15-SP4.

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Mar 22 15:25:03 UTC 2022 - José Iván López González <jlopez@suse.com>
+
+- Fix import users: do not fail if the group does not exist
+  (bsc#1197040).
+- 4.4.11
+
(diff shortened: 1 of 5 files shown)


dgdavid (approved):
LGTM

PR #362, opened 2022-04-08
Fix import users

lslezak:
- Fix import users: do not fail if the group does not exist
- https://bugzilla.suse.com/show_bug.cgi?id=1197040


joseivanlopez (approved):
LGTM

PR #368, opened 2022-08-19
Fix ssh keys nohome

jreidinger:
## Problem

When in autoyast profile home path is not specied, then ssh keys are not written.

- https://bugzilla.suse.com/show_bug.cgi?id=1201185

## Solution

Read system users again before writting ssh keys to know its home paths.

## Testing

- *Unit tests adapted*
- *Tested manually*

### DIFF
--- a/src/lib/y2users/linux/users_writer.rb
+++ b/src/lib/y2users/linux/users_writer.rb
@@ -116,7 +116,9 @@ module Y2Users
         @users_to_write_ssh_keys.each_pair do |user, old_keys|
           system_user = system_users.by_name(user.name)
           if !system_user
-            issues << Y2Issues::Issue.new(_("Failed to find user with name '#{user.name}'"))
+            issues << Y2Issues::Issue.new(
+              format(_("Failed to find user with name '%s'"), user.name)
+            )
             log.error("Failed to find user with name #{user.name}")
             next
           end
--- a/src/lib/y2users/linux/users_writer.rb
+++ b/src/lib/y2users/linux/users_writer.rb
@@ -121,6 +121,7 @@ module Y2Users
             next
           end
 
+          system_user.authorized_keys = user.authorized_keys
           write_user_auth_keys(system_user, old_keys)
         end
       end
@@ -168,9 +169,9 @@ module Y2Users
         edit_password(target_user) if initial_user.password != target_user.password
 
         previous_keys = initial_user.authorized_keys || []
-        if previous_keys != target_user.authorized_keys
-          @users_to_write_ssh_keys[target_user] = previous_keys
-        end
+        return if previous_keys == target_user.authorized_keys
+
+        @users_to_write_ssh_keys[target_user] = previous_keys
       end
 
       # Updates root aliases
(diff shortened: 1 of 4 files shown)


src/lib/y2users/linux/users_writer.rb
```diff
@@ -293,9 +319,7 @@ def adapt_home_ownership(user)
       # @param user [User]
       # @param previous_keys [Array<String>] previous auth keys for given user, if any
       # @return [Boolean] true on success
-      def write_auth_keys(user, previous_keys = [])
-        return true unless exist_user_home?(user)
```
joseivanlopez, line 297: Why isn't home checked? Would it fail if the user was created without home?

joseivanlopez: I see, an issue is generated when `Yast::Users::SSHAuthorizedKeyring::PathError` is raised.

test/lib/y2users/linux/users_writer_test.rb
```diff
@@ -53,6 +53,12 @@
 
   let(:commit_config) { Y2Users::CommitConfig.new }
 
+  let(:system_config) { initial_config }
+
+  before do
+    allow(Y2Users::Linux::Reader).to receive(:new).and_return(double(read: system_config))
```
joseivanlopez, line 59: NP: I would use `instance_double`

test/lib/y2users/linux/users_writer_test.rb
```diff
@@ -331,56 +337,44 @@ def issues(messages)
...
 
         context "and the authorized keys has changed" do
+          let(:system_user) { target_user.copy }
+
           before do
             target_user.authorized_keys = ["new-key"]
+            # cannot overwrite here system_config as we need to have there recent
+            # target user which is modified in `before` code
+            system_config = Y2Users::Config.new.tap { |c| c.attach([system_user]) }
+            allow(Y2Users::Linux::Reader).to receive(:new).and_return(double(read: system_config))
```
joseivanlopez, line 347: And here too.

test/lib/y2users/linux/users_writer_test.rb
```diff
@@ -331,56 +337,44 @@ def issues(messages)
         end
 
         context "and the authorized keys has changed" do
+          let(:system_user) { target_user.copy }
+
           before do
             target_user.authorized_keys = ["new-key"]
+            # cannot overwrite here system_config as we need to have there recent
+            # target user which is modified in `before` code
+            system_config = Y2Users::Config.new.tap { |c| c.attach([system_user]) }
```
joseivanlopez, line 346: NP: you don't need to pass a list to `#attach`.

src/lib/y2users/linux/users_writer.rb
jreidinger: yes and I also think that on system it is not possible to have user without home. at least man page does not mention it and also my own passwd does not have single entry without home. Even one without real home has something like `/bin` or `/sbin`

joseivanlopez: You are right. Checking the code, I see that the home can be created as empty (without skel) but the directory is always created.

joseivanlopez (approved):
LGTM

PR #371, opened 2022-12-08
Use Fallback for CHARACTER_CLASS

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1205783

## Trello

https://trello.com/c/IEHa2Ffx/

## Problem

When trying to create a new user group on a newly installed Tumbleweed, you always get an error popup about the permitted characters for the group name:

```
The name of the group may only contain letters, digits, 
"-", ".", and "_" and must start with a letter.

Please try again.
```

Even though that group name fullfilled those criteria.

## Cause

The `/usr/etc/login.defs` or `/etc/login.defs` file on Tumbleweed now no longer contains a CHARACTER_CLASS field. The code, however, just overwrote the internal variable holding the contents of that field, and if nothing could be read, it would remain empty; and comparing the group name that the user entered against that always failed.

## Fix

Handle that case gracefully: Fall back to a predefined character class if it can't be read from file.

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Dec  8 15:26:35 UTC 2022 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Use a fallback if CHARACTER_CLASS cannot be read from /etc/login.defs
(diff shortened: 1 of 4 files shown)


lslezak (approved):
LGTM

PR #374, opened 2023-01-14
Support for alternative YPPWD_SRCDIR

ancorgs:
## Problem

In a NIS master server, there is a setting stored at `/etc/sysconfig/ypserv` and called `YPPWD_SRCDIR` that specifies where are the `passwd`, `shadow` and `group` files that NIS should export. By default this points to `/etc`. So in a typical NIS master server, managing the users and groups of the NIS domain is fully equivalent to managing the local users and groups in the system.

But what happens if that setting is set to a non default value?

When YaST Users runs in a system that is a NIS master server with `YPPWD_SRCDIR` pointing to an alternative location, it allows the user to select which set of "local users" should be managed in that execution of YaST - the one stored at `/etc` or the one stored at the location pointed by `YPPWD_SRCDIR`.

01_start_yast2_users

Unfortunately, when rewriting YaST Users to rely on the `shadow` tools (`useradd` and friends) we were not aware of that circumstance and YaST always modifies the files under `/etc`.

See https://bugzilla.suse.com/show_bug.cgi?id=1206627

## Solution

First step. Make `Linux::Writer` more configurable so it can be instructed to write the changes to the files in the directory pointed by `YPPWD_SRCDIR`

Second step. Modify `Linux::Writer` and its action to execute the corresponding commands using the `--prefix` argument (for all commands `useradd`, `userdel`, `usermod`, `groupadd`, `groupdel`, `groupmod`, `passwd`, `chpasswd` and `chage`).

## Note for reviewer

This involved a previous refactoring to make easier to add new values to the `Linux::Writer` configuration. Review commit by commit if you want to easily follow the rationale of that refactoring.

## Testing

- Tested manually.
- Added a few unit tests.
- Also verified by the L3 reporter (incident closed).

### DIFF
--- a/src/lib/y2users/linux/root_path.rb
+++ b/src/lib/y2users/linux/root_path.rb
@@ -26,10 +26,9 @@ module Y2Users
     # allow to define the path to what would be the root directory of a system. That implies the
     # tools always assume an extra ./etc directory within the provided path.
     #
-    # Some shadow commands provide both a --prefix argument and a --root one to specify such a root
-    # directory. Other commands provide only the --root one. Since --prefix is more convenient for
-    # the YaST purposes, that will be used by default unless another one is configured for the class
-    # using the provided macro root_path_option.
+    # The shadow commands provide both a --prefix argument and a --root one to specify such a root
+    # directory. YaST uses --prefix because the argument --root assumes the root directory contains
+    # a full system in which the commands may even try to authenticate.
     #
     # For more information, see bsc#1206627
     #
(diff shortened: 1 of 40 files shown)


joseivanlopez (commented):
Great improvement! Thanks.

src/lib/y2users/linux/root_path.rb
```diff
@@ -0,0 +1,56 @@
...
+    # a full system in which the commands may even try to authenticate.
+    #
+    # For more information, see bsc#1206627
+    #
+    # The user of the mixin is expected to set its instance variable @root_path to a non-null value
+    # in order to indicate that the commands should indeed act on that non-default location.
+    #
+    # Hopefully, this mixin will disappear (maybe substituted by a similar one) once the shadow
+    # tools gain the ability to specify the exact location of the passwd, shadow and groups files.
+    module RootPath
```
joseivanlopez, line 40: NP: Maybe this mixin can be part of the `Action` base class.

joseivanlopez (approved):
LGTM

CI failing, but the failure is unrelated to the changes in this PR.

jubalh:
>  I will not approve it untill having a solution from the shadow tools for that commands missing --prefix option.

The shadow PR got merged today.
Upstream plans to create a new release soonish.

joseivanlopez (approved):
LGTM

PR #377, opened 2023-04-03
[SP5] Stop mangling btrfs_subvolume as boolean at Users.EditUser

ancorgs:
Merge of #376 (which fixes [bsc#1209377](https://bugzilla.suse.com/show_bug.cgi?id=1209377)), into the SLE-15-SP5 branch.

While doing so, I found that #370 was fixed only in SLE-15-SP4 and never merged to more modern branches. Since @joseivanlopez cannot remember a good reason for that, we guess it was simply overlooked. So I'm taking the opportunity to merge also that pull request (which fixed [bsc#1202974](https://bugzilla.suse.com/show_bug.cgi?id=1202974)).


joseivanlopez (approved):
LGTM

PR #379, opened 2023-05-22
Support for alternative YPPWD_SRCDIR (alt version)

ancorgs:
## Problem

In a NIS master server, there is a setting stored at `/etc/sysconfig/ypserv` and called `YPPWD_SRCDIR` that specifies where are the `passwd`, `shadow` and `group` files that NIS should export. By default this points to `/etc`. So in a typical NIS master server, managing the users and groups of the NIS domain is fully equivalent to managing the local users and groups in the system.

But what happens if that setting is set to a non default value?

When YaST Users runs in a system that is a NIS master server with `YPPWD_SRCDIR` pointing to an alternative location, it allows the user to select which set of "local users" should be managed in that execution of YaST - the one stored at `/etc` or the one stored at the location pointed by `YPPWD_SRCDIR`.

01_start_yast2_users

Unfortunately, when rewriting YaST Users to rely on the `shadow` tools (`useradd` and friends) we were not aware of that circumstance and YaST always modifies the files under `/etc`.

See https://bugzilla.suse.com/show_bug.cgi?id=1206627

## Solution

First step. Make `Linux::Writer` more configurable so it can be instructed to write the changes to the files in the directory pointed by `YPPWD_SRCDIR`

Second step. Modify `Linux::Writer` and its action to execute the corresponding commands using the `--prefix` argument (for `useradd`, `userdel`, `usermod`, `groupadd`, `groupdel`, `groupmod`) or the `--root` one (for `passwd`, `chpasswd`, `chage`).

## Note for reviewer

This involved a previous refactoring to make easier to add new values to the `Linux::Writer` configuration. Review commit by commit if you want to easily follow the rationale of that refactoring.

## Testing

Still pending.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -109,8 +109,8 @@ ylib_y2users_DATA = \
   lib/y2users/groups_collection.rb \
   lib/y2users/read_result.rb \
   lib/y2users/home.rb \
-  lib/y2users/commit_config.rb \
-  lib/y2users/commit_config_collection.rb
+  lib/y2users/user_commit_config.rb \
+  lib/y2users/user_commit_config_collection.rb
 
 ylib_y2users_clientsdir = @ylibdir@/y2users/clients
 ylib_y2users_clients_DATA = \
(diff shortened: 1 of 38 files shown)


ancorgs:
This was created just because I had problems updating  #374. Now the problems are fixed and this is not needed anymore.

PR #380, opened 2023-06-08
Write users when running AutoYaST on an installed system

imobachgs:
## Problem

* [bsc#1211753](https://bugzilla.suse.com/show_bug.cgi?id=1211753)

Users are not written when using AutoYaST on an already installed system, neither using `ayast_setup` nor the AutoYaST UI. In those scenarios, the process of detecting which users/groups changed is tricky:
                                                                                                                                                                                         
* The approach followed by [Y2Users::UsersModule::Reader](https://github.com/yast/yast-users/blob/414b6c7373068c367c0a01be20a1399fbd0ef470/src/lib/y2users/users_module/reader.rb#L103), checking the content of `org_user`, does not work because it is defined only if the user was modified using the AutoYaST UI.
* Directly comparing the users/groups from `system_config` and `target_config` does not work because passwords are missing from the `target_config` users.

```
{
  :class=>Y2Users::User, :name=>"systemd-timesync", :uid=>"496", :gid=>"496", :shell=>"/usr/sbin/nologin",
  :home=>#<Y2Users::Home:0x00005585a57372d0 @path="/", @permissions=nil, @btrfs_subvol=nil>,
  :gecos=>["systemd Time Synchronization"], :source=>:unknown, :password=>nil,
  :authorized_keys=>[], :receive_system_mail=>false, :secondary_groups_name=>[]
}
```

vs

```
{
  :class=>Y2Users::User, :name=>"systemd-timesync", :uid=>"496", :gid=>"496", :shell=>"/usr/sbin/nologin",
  :home=>#<Y2Users::Home:0x00005585a7bb4e78 @path="/">,
  :gecos=>["systemd Time Synchronization"], :source=>:unknown,
  :password=>#<Y2Users::Password:0x00005585a79f6730 @value=#Y2Users::PasswordEncryptedValue:0x00005585a79f67f8 @content=<secret>>, @aging=#<Y2Users::PasswordAging:0x00005585a79f66e0 @content="19515">, @minimum_age="", @maximum_age="", @warning_period="", @inactivity_period="", @account_expiration=#<Y2Users::AccountExpiration:0x00005585a79f66b8 @content="">>,
  :authorized_keys=>[], :receive_system_mail=>nil, :secondary_groups_name=>[]
}
```

## Solution
                                                                                                                                                                                      
To overcome these limitations, we only consider those users who 'modified'
property is not nil, although it does not guarantees that the user/group has
changed at all.

## Testing

- [x] Added a new unit test*
- Manually tested scenarios:
  - [x] AutoYaST UI importing a profile
  - [x] AutoYaST UI reading the users from the system (cloning)
  - [x] `yast2 clone_system`
  - [x] `yast2 ayast_setup setup filename=users.xml` (adding)
  - [x] `yast2 ayast_setup setup filename=users.xml` (editing a user)

### DIFF
--- a/src/lib/users/clients/auto.rb
+++ b/src/lib/users/clients/auto.rb
@@ -22,6 +22,10 @@ require "installation/auto_client"
 require "y2users"
 require "y2users/autoinst/reader"
 require "y2issues"
+require "y2users/config_merger"
+require "y2users/config_manager"
+require "y2users/autoinst/reader"
+require "y2users/linux/writer"
 
 Yast.import "Users"
 Yast.import "Linuxrc"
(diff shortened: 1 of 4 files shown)


src/lib/users/clients/auto.rb
```diff
@@ -90,15 +94,33 @@ def read
...
+      # To overcome these limitations, we only consider those users/groups
+      # which 'modified' property is not nil, although it does not guarantee
+      # that they changed at all.
       #
       # @return [Boolean] true if configuration was changed; false otherwise.
       def write
-        Yast::Users.SetWriteOnly(true)
-        progress_orig = Yast::Progress.set(false)
-        ret = Yast::Users.Write == ""
-        Yast::Progress.set(progress_orig)
```
ancorgs, line 100: Progress handling is gone in the new code? Was it unnecessary?

imobachgs: That's a good question. `Yast::Users.Write` uses the `Progress` API. However, the new code does not use it (or at least that's what I expect), so I do not think it is needed anymore.

ancorgs: Fine then. I was just curious.

PR #383, opened 2023-06-30
User creation page got messed when going back after importing user(s) from existing installation

mchf:
## Problem

Reporter imported a user from existing system during installation. Later on walked back and the user creation dialog's password field was messed with incredibly long password. Most probably hashed password got in use.

- bsc#1211583

## Solution

Dialog tries to determine whether a user was already imported when walking back. If yes it disables manual definition part of the dialog and enables importing user

## Testing

- Tested manually

TODO: automatic tests


ancorgs (approved):
LGTM

PR #391, opened 2023-09-25
Don't use Obsolete Method `Dir.exists?`

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1215637

## Problem

Crash when trying to use the directory selector in the "Edit User" or "Add User" dialog with _"undefined method `exists?' for Dir:Class"_.

## Cause

`Dir.exists?` and `File.exists?` were deprecated long ago with Ruby 2.2, and now with Ruby 3.2 it was finally removed.

Both methods were just aliases for `Dir.exist?` and `File.exist?`, respectively (even though that's the usual Ruby slightly broken English).

## Fix

Now using `Dir.exist?` instead.

## What about Other Releases like SLE-15.x / Leap 15.x?

Those are still using Ruby 2.5 (SLE-15.5 / Leap 15.5) or older. No need to change it there as well.

## Test

Manual test on TW.

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Sep 25 12:33:45 UTC 2023 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Don't use obsolete method Dir.exists? (bsc#1215637)
+- 5.0.1 
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


shundhammer:
Manually submitted to `openSUSE:Factory`:

https://build.opensuse.org/request/show/1113640

PR #392, opened 2023-10-31
Drop obsolete comment documenting a feature that does not exist

Dzejrou:
## Problem

While working on bsc#1216518 [*] where the SUSE docs implied the feature described by this comment existed I encountered this comment that confirmed the implication made by the docs. The docs should be hopefully fixed in the future but the existence of this comment is confusing as it implies this (now?) non-existent feature does exist.

[*] Not relevant to the comment itself so I did not mention the bug in the commit.

## Solution

Remove the part of the documenting comment that talks about the feature (random number being generated when '!' is used in the AutoYaST profile with the password declared as not encrypted).

Note: I also removed the second sentence referencing to the `encrypted` attribute. I did this because that attribute affects how the '!' password is interpreted and the reference was immediately after the removed part (rather than after the description of the `user_password` attribute) so I assumed the reference existed because of the random password feature. If it should remain, let me know and I will amend the commit.

### DIFF
--- a/src/lib/y2users/autoinst_profile/user_section.rb
+++ b/src/lib/y2users/autoinst_profile/user_section.rb
@@ -89,8 +89,7 @@ module Y2Users
       #   @return [String,nil] Default shell
       #
       # @!attribute user_password
-      #   @return [String,nil] User's password. If set to an exclamation mark ('!'),
-      #   a random password is generated. See #encrypted.
+      #   @return [String,nil] User's password.
       #
       # @!attribute encrypted
       #   @return [Boolean,nil] Determine whether #user_password is encrypted or not.


src/lib/y2users/autoinst_profile/user_section.rb
```diff
@@ -89,8 +89,7 @@ def self.attributes
       #   @return [String,nil] Default shell
       #
       # @!attribute user_password
-      #   @return [String,nil] User's password. If set to an exclamation mark ('!'),
-      #   a random password is generated. See #encrypted.
+      #   @return [String,nil] User's password.
```
jreidinger, line 92: yeap, it is not random, but it disables login with password. Quotation from man page for shadow:
```
If the password field contains some string that is not a valid result of crypt(3), for instance ! or *, the user will not be able to use a unix password to log in (but the
           user may log in the system by other means).
```

jreidinger (approved):

PR #395, opened 2024-08-20
Relax Check in GECOS Field: Allow any Data except Colons [SLE-15-SP6]

shundhammer:
## Target Branch

This is for **SLE-15-SP6**. A merge to Factory / TW will follow.

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1228149

## Trello

https://trello.com/c/PLxx2iDV

## Problem

PAM in connection with some desktops like GNOME need an _umask_ specification to be set in the GECOS field (the comment / full user name field) in `/etc/passwd`.

But the YaST users module complains that only 3 fields separated by commas are permitted:

y-users-gecos-error

This is triggered when trying to leave that tab.

## Cause

There is an additional check in the ancient `Users.pm` Perl module that dates back to 2004. Why this check exists is unclear; there doesn't seem to be a reference why this limitation was introduced.

## Fix

Removed that check.

## Test

Manual test in a TW VM. After removing this check, it doesn't complain anymore, and `/etc/password` now contains the expected entry:

```console
% grep '^kilroy' /etc/passwd

kilroy:x:1001:100:Kilroy Anon,,,,UMASK=022:/home/kilroy:/bin/bash
```

## Further Reading

- `man 1 chfn`
- `man 8 pam_umask`

## Related PR

- Merge to Factory / TW:  _TBD_

### DIFF
--- a/package/yast2-users.changes
+++ b/package/yast2-users.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Aug 21 13:21:10 UTC 2024 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Relax check in GECOS field (bsc#1228149):
+  Allow any data except colons
+- 4.6.6
+
+-------------------------------------------------------------------
 Wed Aug 21 12:33:48 UTC 2024 - Josef Reidinger <jreidinger@suse.com>
 
 - Backport changes to avoid namespace collisions
--- a/package/yast2-users.spec
+++ b/package/yast2-users.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-users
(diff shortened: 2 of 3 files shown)


mvidner:
Archaeology time! :)
The check was added in commit 511d3d0dd52f32587eeb81b3d37a728a0c3c6d9a in 2003, in response bug suse18270 which is https://bugzilla.suse.com/show_bug.cgi?id=33270
At that time, the passwd(5) man page probably contained some details about GECOS, but there was no specific bug.
Funnily, even then we knew about `umask=`.

shundhammer:
See also #396

jreidinger (approved):

PR #401, opened 2025-07-04
Temporarily disable Users#Export tests"

imobachgs:
- Temporarly disable Users#Export tests
- Bump version and update changes file

See [bsc#1245519](https://bugzilla.suse.com/show_bug.cgi?id=1245519) for further details as it is not used in SLFO anyways.


ancorgs (approved):
LGTM. Not the optimal solution but we are...

https://youtu.be/a01QQZyl-_I

